Issue #34, opened 2020-08-10
Upgrade EC terraform provider to use latest sdk provider functions


ggsood:
## Overview
Upgrade EC terraform provider to use latest SDKs provider functions introduced as part of SDK2 , which adds diagnosis capabilities during provider initialisation.

## Possible Implementation
Refactor the provider code according to the new SDK schema.

## Testing
the provider works as per the current implementation.

## Context
Using the latest features provided by the sdk2 kit of terrafom. 

## Your Environment
Mac OS-X
Golang version 1.14.6

marclop:
Resolved by #43

PR #35, opened 2020-08-10
provider: Use ConfigureContextFunc

ggsood:
EC Provider refactoring to use new ConfigureContextFunc

## Description
The change will refactor the EC provider code to use new ConfigureContextFunc which provides better diagnosis capabilities.

## Related Issues
https://github.com/elastic/terraform-provider-ec/issues/34

## Motivation and Context
Use the updated terraform sdk functionalities.

## How Has This Been Tested?
Locally by consuming the new provider and creating resources at Elastic cloud.

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/ec/provider_config.go
+++ b/ec/provider_config.go
@@ -20,7 +20,6 @@ package ec
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"net/http"
 	"os"
 	"time"
@@ -28,6 +27,7 @@ import (
 	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/auth"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 )
 
 const (
--- a/ec/provider_config.go
+++ b/ec/provider_config.go
@@ -26,8 +26,8 @@ import (
 
 	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/auth"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 const (
--- a/ec/provider_config.go
+++ b/ec/provider_config.go
@@ -18,7 +18,9 @@
 package ec
 
 import (
+	"context"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"net/http"
 	"os"
 	"time"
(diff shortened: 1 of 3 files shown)


ec/provider_config.go
```diff
@@ -33,22 +35,36 @@ const (
...
+	// Warning or errors can be collected in a slice type
+	var diags diag.Diagnostics
 	if err != nil {
-		return nil, err
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "Unable to get EC client",
+			Detail:   "Invalid timeout value for EC client",
+		})
+		return nil, diags
```
karencfv, line 49: It's a good idea to return the errors that we are getting from other functions as they are. This way, what went wrong is clearer. Let's use the `FromErr` function instead :) 
```suggestion
		return nil, diag.FromErr(err)
```

ec/provider_config.go
```diff
@@ -33,22 +35,36 @@ const (
...
 	})
+
 	if err != nil {
-		return nil, err
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "Unable to get EC client",
+			Detail:   "Invalid credentials for EC client",
+		})
+		return nil, diags
```
karencfv, line 64:
```suggestion
		return nil, diag.FromErr(err)
```

ec/provider_config.go
```diff
@@ -58,6 +74,18 @@ func configureAPI(d *schema.ResourceData) (interface{}, error) {
...
 	})
+
+	if err != nil {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "Unable to create EC client",
+			Detail:   "Unable to authenticate user for EC client",
+		})
+
+		return nil, diags
```
karencfv, line 85:
```suggestion
		return nil, diag.FromErr(err)
```

ec/provider_config.go
```diff
@@ -58,6 +74,18 @@ func configureAPI(d *schema.ResourceData) (interface{}, error) {
...
+	}
+
+	return api, diags
```
karencfv, line 88: We can return `nil` directly here as all errors are fatal and we are not collecting a series of diagnostics yet.
```suggestion
	return api, nil
```

ec/provider_config.go
```diff
@@ -33,22 +35,36 @@ const (
 )
 
 // configureAPI implements schema.ConfigureFunc
-func configureAPI(d *schema.ResourceData) (interface{}, error) {
+func configureAPI(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
 	timeout, err := time.ParseDuration(d.Get("timeout").(string))
+
+	// Warning or errors can be collected in a slice type
+	var diags diag.Diagnostics
```
karencfv, line 42: This won't be necessary any more if we use the `FromErr` function, and we are not building a collection of several diagnostics

ec/provider_config.go
ggsood: Thanks @karencfv for the review 😄 have addressed the suggested changes.

ec/provider_config.go
```diff
@@ -18,7 +18,9 @@
 package ec
 
 import (
+	"context"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
marclop, line 23: This import should be moved to line 28, and preserve the space between the stdlib imports and the external ones. It's something that we should have in our linter. I'll add that linter check so this doesn't happen again.

ggsood: Thanks @marclop for the review, have corrected the order of imports.

marclop:
running `make format` should fix the imports and produce this patch:

```diff
diff --git a/ec/provider_config.go b/ec/provider_config.go
index 2ba3524..4d28885 100644
--- a/ec/provider_config.go
+++ b/ec/provider_config.go
@@ -26,8 +26,8 @@ import (
 
 	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/auth"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 const (

```

ggsood:
Thanks @marclop have fixed the lint errors.

PR #40, opened 2020-08-11
ci: Add Acceptance tests to GitHub action workflow

marclop:
## Description

Adds the acceptance tests `make testacc` run to the workflow run in all
the commits and branches, but omitting the pull_request trigger due to
the GitHub Actions limitations on forks.

## Motivation and Context

Run the acceptance tests in tests in GH Actions.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -1,11 +1,11 @@
 name: Go
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
-  # Unit test run
-  unit:
-    name: Unit
+  # Acceptance tests run.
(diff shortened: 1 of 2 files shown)


marclop:
that's a good point, I actually had the trigger first to both Unit and Acceptance to run on both `push` and `pull_request` triggers but the Action secrets have a limitation which is described in https://github.com/elastic/terraform-provider-ec/settings/secrets.

```
Secrets are environment variables that are encrypted and only exposed to selected actions. Anyone with collaborator access to this repository can use these secrets in a workflow.

Secrets are not passed to workflows that are triggered by a pull request from a fork. Learn more.
```

PR #41, opened 2020-08-13
ec_deployment: Add APM config setting management

### LINKED ISSUE #7: ec_deployment: Add settings object to apm resource
## Description

Currently, there's no way of managing an Apm resource settings, this is not desirable as users might want to control which settings their Apm deployment resources have set.

There's quite a lot of fields and sub-objects to implement, so this might take some time to implement.

marclop:
## Description
This patch is split in different commits to help review the changes more
easily.

Mainly, it adds a `config {}` block to the `ec_deployment.apm` resource
both at the top level `ec_deployment.<name>.apm[0].config` and at the
topology level `ec_deployment.<name>.apm[0].topology[0].config` to match
the override capacity of the API.

The top level object provides settings for all the topology elements and
the topology element config object overrides settings for that element.

All fields are optional with no default value, there are two exceptions:

* config.debug_enabled is optional and defaults to `false`.
* config.secret_token is computed AND optional, meaning that the value
can be overridden.

Additionally, some clean up and reorg has gone into this patch, namely:

* Split schema into different files, each file represents a deployment
  resource.
* The previous credentials for elasticsearch have been moved to the top
  level of the deployment resource as `elasticsearch_username` and
  `elasticsearch_password`.
* The APM `secret_token` parsing has been removed as it's present in the
  apm resource config and can be used as `apm[0].config.secret_token`.
* Updates the `example/deployment` version and deployment template id.
* Reorganizes the acceptance test folder to have a `deployment_basic.go`
  file with the acceptance test case. Additionally, two new `.tf` files
  have been added to cover the APM `config` block.
* Last, adds a new `TEST_NAME` makefile variable which allows users to
  target a single test case in `make testacc`.

## Related Issues
Resolves #7 

## How Has This Been Tested?

`make testacc`

## Types of Changes


ec/ecresource/deploymentresource/apmstate/apm_expanders.go
```diff
@@ -106,8 +116,81 @@ func expandTopology(raw interface{}) ([]*models.ApmTopologyElement, error) {
...
+	}
+
+	if !reflect.DeepEqual(res, new(models.ApmConfiguration)) {
+		return res
+	}
+
+	return nil
+}
+
+func initSystemSettings(cfg *models.ApmConfiguration) {
```
karencfv, line 192: Why does this function need to be called so many times within `expandConfig` ?

ec/acc/deployment_test.go
```diff
@@ -22,65 +22,10 @@ import (
...
-	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-const (
-	deploymentVersion          = "7.8.1"
-	deploymentVersionAppsearch = "7.6.2"
-	region                     = "us-east-1"
-)
-
-func TestAccDeployment_basic(t *testing.T) {
```
karencfv, line 38: If this test was extracted into a separate file, should the other tests be extracted too?

ec/ecresource/deploymentresource/schema.go
```diff
@@ -43,496 +45,56 @@ func NewSchema() map[string]*schema.Schema {
...
-				},
-			},
-		},
-	}
+// suppressMissingOptionalConfigurationBlock handles configuration block attributes in the following scenario:
+//  * The resource schema includes an optional configuration block with defaults
+//  * The API response includes those defaults to refresh into the Terraform state
+//  * The operator's configuration omits the optional configuration block
+func suppressMissingOptionalConfigurationBlock(k, old, new string, d *schema.ResourceData) bool {
+	return old == "1" && new == "0"
```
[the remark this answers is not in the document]
marclop: this will only happen when the block element `apm[0].config.#` goes from 1 to 0, meaning that the number of items in the TypeList goes from 1 to 0.

ec/acc/deployment_test.go
marclop: yeah, but it's out of the scope of this PR

ec/ecresource/deploymentresource/schema_resource_settings.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func userSettingsOverrideYAML() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeString,
+		Description: `An arbitrary YAML object allowing (non-admin) cluster owners to set their parameters (only one of this and 'user_settings_json' is allowed), provided they are on the whitelist ('user_settings_whitelist') and not on the blacklist ('user_settings_blacklist'). (These field together with 'user_settings_override*' and 'system_settings' defines the total set of resource settings)`,
+		Optional:    true,
+	}
+}
```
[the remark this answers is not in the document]
marclop: that makes sense, I just did this in case it made sense, but I do agree that it makes the schemas somewhat less readable

ec/ecresource/deploymentresource/apmstate/apm_expanders.go
marclop: I'll address this, it isn't necessary anymore

marclop: b244157 changes the behaviour to match the defaults in the schema, there's no case where the `SystemSettings` would be empty now, so it doesn't make sense to add the complexity to tackle that

PR #45, opened 2020-08-17
terraform: Add support for 0.13 version

marclop:
## Description

Adds support for terraform 0.13, which introduces a few changes on how
third party providers need to be declared in the `.tf` files. Namely, it
modifies the `make install` target so that the binary is placed both on
the new directory structure and the old one at the top level.

Additionally, it modifies the examples and the example in the README to
include the `elastic/ec` source.

If the `required_providers` block is not specified, terraform v0.13 will
attempt to download the terraform binary from:
`registry.terraform.io/hashicorp/ec`.

## Motivation and Context

Support terraform `0.13+`

## Types of Changes

- [x] Documentation


examples/deployment_appsearch/deployment.tf
```diff
@@ -1,5 +1,10 @@
...
-  required_version = ">= 0.12"
+  required_version = ">= 0.12.29"
+  required_providers {
+    ec = {
+      source = "elastic/ec"
+    }
+  }
 }
 
 provider "ec" {
```
rhass, line 10: Minor nit, `fmt` changes this:

```
-provider "ec" {
-}
+provider "ec" {}
```

examples/deployment_appsearch/deployment.tf
```diff
@@ -1,5 +1,10 @@
 terraform {
-  required_version = ">= 0.12"
+  required_version = ">= 0.12.29"
+  required_providers {
```
rhass, line 3: `fmt` injects a newline break here:

```
   required_version = ">= 0.12.29"
+
   required_providers {
```

PR #46, opened 2020-08-17
ci: Add validate-examples target in action run

marclop:
## Description

Adds the `validate-examples` makefile target which runs `terraform init`
and `terraform validate` in each of the committed examples.

## Motivation and Context

Validate the committed terraform examples in the CI folder.

## How Has This Been Tested?

`make validate-examples`


.github/workflows/go.yml
```diff
@@ -18,6 +18,11 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Install terraform
+    - uses: hashicorp/setup-terraform@v1
+      with:
+        terraform_version: 0.12.29
```
ggsood, line 24: Can upgrade to terraform 0.13

marclop: After #45 is merged

PR #50, opened 2020-08-20
acc: Remove deployment.Healthy check

### LINKED ISSUE #8: ec_deployment: Add data source
## Description

Implement a data resource which allows the user to search deployments, and retrieve a result.

The deployment API has the Search API available which uses `models.SearchRequest` and contains a `models.QueryContainer`, giving the user A LOT of options and queries.

I wonder if we should have a simple and an advanced querying mechanism, with simplified queries for users.

```hcl
data "ec_deployment" "example" {
  # Based on models.QueryContainer
  query {
    nested {
      path = "plan_info.current.plan.cluster_topology"
      query {
        bool {
        }
      }
    }
  }
}
```

## Reference

I believe the `aws_iam_policy_document` is a good resource for inspiration: <https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html>

### Terraform data resource reference

<https://www.terraform.io/docs/configuration/data-sources.html>

### LINKED ISSUE #38: acceptance: Tests fail with errors unrelated to TF provider
## Description
Currently acceptance tests can sometimes fail when there are errors 
unrelated to the terraform provider.

For example, if a deployment is unhealthy (unassigned shard or whatever),
or the deployment failed to delete (which only issues a warning), etc, the tests will fail. 

It seems to me that it would probably make sense to either log these errors somehow,
or not take these errors into account at all as they can provide false positives.

marclop:
## Description

Removes the acceptance `deployment.Healthy` since it creates a lot of
instability in the acceptance test results.

## Related Issues

Resolves #38  

## Types of Changes

### DIFF
--- a/ec/acc/deployment_checks_test.go
+++ b/ec/acc/deployment_checks_test.go
@@ -20,6 +20,7 @@ package acc
 import (
 	"fmt"
 
+	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/api/deploymentapi"
 	"github.com/elastic/cloud-sdk-go/pkg/api/deploymentapi/deputil"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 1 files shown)


karencfv (approved):
👌 thank you!

PR #52, opened 2020-08-31
ec_deployment_traffic_filter: Add new resource

### LINKED ISSUE #48: ec_deployment: Add support for traffic filtering
## Overview
I would like support for IP traffic filters.
## Possible Implementation
We would need a new resource "ec_ip_traffic_filter". Next to that, we can have an optional field in "ec_deployment" that can point to a resource of type "ec_ip_traffic_filter".

We will also need a data source "ec_ip_traffic_filter" for when the filter is created externally, and you want to associate a deployment with that filter.

## Context
I have Azure Web applications that need to connect to ES (hosted in EC). Ideally I can dynamically add the x-amount of outbound IP adresses to the the Elasticsearch deployment.

I currently use the same technique to add [whitelist entries](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/project_ip_whitelist) to MongoDB Atlas projects 

## Your Environment
We have different deployments for different environments (dev/staging/prod). I would like to add a different ruleset to each deployment.

When changing the filter, I don't want any connection downtime (I have not tested this yet without Terraform as this is quite a new feature.

marclop:
## Description
Adds a new resource `ec_deployment_traffic_filter` and modifies the
existing `ec_deployment` resource to add a new field `traffic_filter`,
where traffic filter IDs can be specified.

The patch has been split in three different commits

#### Add the `ec_deployment_traffic_filter` resource (35a28d9)

Adds the resource and makes the field `include_by_default` default to
false like we would in our `cloud-sdk-go` when not specified, rather
than a required field, like the API does.

#### Add `traffic_filter` to `ec_deployment` (b760a31)

The new `traffic_filter` is an optional field of type Set with the hash
algorithm "HashString", it's a list of unique strings which reference a
number of traffic_filtering rules.

On the implementation, there's two ways the traffic_filter rules have to
be handled. 
On Creation, theres a `settings` field on the `CreateRequest`, which is
not present on the `UpdateRequest`, Create specifies a list of traffic 
filters by ID which are added to the deployment on creation.
On Update, there's no `settings` field so we need to rely on the traffic
filter associations which the API present.
On Delete, there's also the challenge of deleting the existing traffic
filter associations before the deployment is Shut down.

To be able to handle the Update and Delete, `handleTrafficFilterChange`
is introduced which handles, the addition and the removal of items from
that `traffic_filter` list.

`handleTrafficFilterChange` checks if there's a `traffic_filter` change
and adds and removes associations of a deployment to a traffic filter
rule set.

Direct use of the V1 APIs without a wrapper package is necessary here
since there's no actual implementation of these APIs on the cloud-sdk-go
(yet). A follow-up PR will address this point.

Last, a `hasDeploymentChange` function has been created to only update
the deployment through PUT with the current state when there's a change
in any field != `traffic_filter` this avoids creating unnecessary no-op
plans on the backend which slow down the changes and clutter the plan
history.

#### Add acceptance tests for both resources (4bca40e)

Adds the integration tests which handle the CRUD of the new resource
`ec_deployment_traffic_filter`, the case where a deployment is assigned
new `traffic_filter` rules and `ec_deployment_traffic_filter` sweepers.

## Related Issues

Resolves #48 

## How Has This Been Tested?

By running `make testacc`

## Types of Changes


marclop:
`make testacc` result:

```
$ make testacc
-> Formatting Go files...
-> Formatting Terraform documentation blocks...
-> Done.
=== RUN   TestAccDeployment_appsearch
=== PAUSE TestAccDeployment_appsearch
=== RUN   TestAccDeployment_basic
=== PAUSE TestAccDeployment_basic
=== RUN   TestAccDeploymentTrafficFilter_basic
=== PAUSE TestAccDeploymentTrafficFilter_basic
=== CONT  TestAccDeployment_appsearch
=== CONT  TestAccDeployment_basic
=== CONT  TestAccDeploymentTrafficFilter_basic
--- PASS: TestAccDeploymentTrafficFilter_basic (15.55s)
--- PASS: TestAccDeployment_appsearch (774.74s)
--- PASS: TestAccDeployment_basic (873.08s)
PASS
ok  	github.com/terraform-providers/terraform-provider-ec/ec/acc	874.503s
```

ec/ecresource/deploymentresource/resource.go
```diff
@@ -15,24 +15,23 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package ecresource
+package deploymentresource
```
karencfv, line 18: Looking at other providers (even though they don't have their resources separated into packages like we do) the naming for their functions is like this: [`resourceLocalFIle`](ec/ecresource/deploymentresource/resource.go) or `resourceAwsAmi`. Going forward I think it would make sense to follow this naming shceme and rename this package to `resourcedeployment`. I know it might look weird, but I _think_ we need to be somewhat consistent with the other providers? what do you think? 🤔

ec/ecresource/deploymentresource/resource.go
```diff
@@ -15,24 +15,23 @@
...
-// Deployment returns the ec_deployment resource schema.
-func Deployment() *schema.Resource {
+// Resource returns the ec_deployment resource schema.
+func Resource() *schema.Resource {
 	return &schema.Resource{
-		CreateContext: deploymentresource.Create,
-		ReadContext:   deploymentresource.Read,
-		UpdateContext: deploymentresource.Update,
-		DeleteContext: deploymentresource.Delete,
+		CreateContext: Create,
```
karencfv, line 29: I can't leave a comment where these functions live 😄 , but it seems like they don't need to be exported anymore?

ec/ecresource/trafficfilterresource/create.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+
+	"github.com/elastic/cloud-sdk-go/pkg/api"
+	"github.com/elastic/cloud-sdk-go/pkg/client/deployments_traffic_filter"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Create will create a new deployment traffic filter ruleset
+func Create(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
karencfv, line 30: Looks like this doesn't need to be exported anymore :)

ec/ecresource/trafficfilterresource/delete.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+
+	"github.com/elastic/cloud-sdk-go/pkg/api"
+	"github.com/elastic/cloud-sdk-go/pkg/client/deployments_traffic_filter"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Delete will delete an existing deployment traffic filter ruleset
+func Delete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
karencfv, line 30: Looks like this doesn't need to be exported anymore :)

ec/ecresource/trafficfilterresource/read.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/elastic/cloud-sdk-go/pkg/api"
+	"github.com/elastic/cloud-sdk-go/pkg/client/deployments_traffic_filter"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Read queries the remote deployment traffic filter ruleset state and update
+// the local state.
+func Read(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
karencfv, line 31: Looks like this doesn't need to be exported anymore :)

ec/ecresource/trafficfilterresource/update.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+
+	"github.com/elastic/cloud-sdk-go/pkg/api"
+	"github.com/elastic/cloud-sdk-go/pkg/client/deployments_traffic_filter"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Update will update an existing deployment traffic filter ruleset
+func Update(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
karencfv, line 30: Looks like this doesn't need to be exported anymore :)

karencfv (commented):
Sorry about the way comments show up now! Looks like Github changed the way comments show up when you make them directly on a specific commit 😒 If they don't make sense or whatever I can write them again or have a chat 😅

karencfv:
Ah! one more thing, it'd be good to add some examples :) Looking at other provider's documentations it looks like they have an example per resource :)

ec/ecresource/deploymentresource/resource.go
marclop: I don't have any strong opinions on naming here, if that's what other providers do, we can easily do that

marclop:
> I was under the impression that these were going to be added to the cloud-sdk-go as APIs first, and then those APIs were going to be used here. I might be wrong, but I think we had a chat about that?
> 
> In any case, it's good practice to have consistency within the different resources, as well as having these defined as APIs with validation etc :)

I can hold the Pull Request open until the APIs have an implementation on the `cloud-sdk-go`, since all these APIs aren't implemented, I instead, chose to used them here. My idea was to merge this first, and then migrate over to the implemented APIs, since we're not shipping code as is and it wouldn't change anything for a user.
There's a bug on the spec which causes the `201` response to be returned as an error, so it's probably not a bad idea to update the SDK when that's fixed and when the traffic filtering APIs are implemented, or do those two separately.

> I'm on the fence about having this be a subpackage for `deploymentresource`. I can't really think of why these functions can't be a part of `deploymentresource`?

I can move that a level up, I just thought I'd keep the deployment package less cluttered, since there's already a good amount of files.

> Ah! one more thing, it'd be good to add some documentation with examples :) Looking at other provider's docs it looks like they have an example per resource :)

I can add the documentation and examples too.

> I can't leave a comment where these functions live 😄 , but it seems like they don't need to be exported anymore?

We can make them private easily, I just forgot 🙂

karencfv:
Thanks for opening up the PR with the traffic filter APIs! much appreciated 🙇‍♀️ 

> I can move that a level up, I just thought I'd keep the deployment package less cluttered, since there's already a good amount of files.
Yeah, I wasn't too sure about that myself 🤔  But in that case it seemed to me that those should be part of the deployment package. It's not too important so I'll leave that up to you to decide :)

karencfv (approved):
👌 ✨

ec/ecresource/deploymentresource/resource.go
karencfv: I don't have super strong opinions either, but just wondering if this might be something that they'll insist on eventually 😄 I've left the repo approved in any case, so you can either change it or not

karencfv:
Ah! one more thing, what about this comment? -> https://github.com/elastic/terraform-provider-ec/pull/52#commitcomment-41925929

marclop:
> Ah! one more thing, it'd be good to add some documentation with examples :) Looking at other provider's docs it looks like they have an example per resource :)

Ah! I meant to create the GitHub issue that we'll be running the acceptance tests on a PR level with the current Elastic Infra CI. Much like APM and Elasticsearch do. It's a pretty good way to run the tests and tests will run automatically for Elastic employees but only run if we trigger them on external contributors, to prevent Secret leakage, and other potential issues

karencfv:
sounds good!

PR #54, opened 2020-09-04
datasource: Implement new ec_deployment datasource

karencfv:
## Description
Implements a new `ec_deployment` data resource, which retrieves information
from an existing deployment.

This information can also be used in conjunction with a resource like the 
following example:
```hcl
data "ec_deployment" "my_deployment" {
  deployment_id = "a8f22a9b9e684a7f94a89df74aa14331"
}

resource "ec_deployment" "example_minimal" {
  name = "my_example_deployment"

  version                = "7.8.1"
  deployment_template_id = "data.ec_deployment.my_deployment.deployment_template_id"

  elasticsearch {
    topology {
      instance_configuration_id = "data.ec_deployment.my_deployment.elasticsearch.0.instance_configuration_id"
      memory_per_node = "data.ec_deployment.my_deployment.elasticsearch.0.memory_per_node"
    }
  }

  kibana {
    topology {
      instance_configuration_id = "data.ec_deployment.my_deployment.kibana.0.instance_configuration_id"
      memory_per_node = "data.ec_deployment.my_deployment.kibana.0.memory_per_node"
    }
  }
}
```

Some parsing functions have also been moved to a new `util` package so
they can be used in all packages.

The second part of the issue, a second datasource which returns deployments
based on a query, will be on a separate PR in order to keep a tight scope and
for ease of reviewing.

### TODO

In order to start receiving some feedback for when I return on Monday, I am 
opening the pull request now. The following is still pending:

- [x] Tests
- [x] Documentation

## Related Issues
Part 1/2 of https://github.com/elastic/terraform-provider-ec/issues/8

## How Has This Been Tested?
Manually
 
## Types of Changes

## Readiness Checklist


ec/ecdatasource/deploymentdatasource/state/flatteners_apm.go
```diff
@@ -0,0 +1,75 @@
...
+		var m = make(map[string]interface{})
+
+		m["healthy"] = *res.Info.Healthy
+
+		m["ref_id"] = *res.RefID
+
+		m["resource_id"] = *res.Info.ID
+
+		var plan = res.Info.PlanInfo.Current.Plan
+		m["version"] = plan.Apm.Version
```
marclop, line 40: `Apm` could be empty if a deployment has been stopped.

ec/ecdatasource/deploymentdatasource/state/flatteners_apm.go
```diff
@@ -0,0 +1,75 @@
...
+
+		m["topology"] = flattenApmTopology(plan)
+
+		m["elasticsearch_cluster_ref_id"] = *res.ElasticsearchClusterRefID
+
+		for k, v := range util.FlattenClusterEndpoint(res.Info.Metadata) {
+			m[k] = v
+		}
+
+		m["secret_token"] = plan.Apm.SystemSettings.SecretToken
```
marclop, line 50: Same with `SystemSettings`.

ec/ecdatasource/deploymentdatasource/state/flatteners_apm.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func flattenApmTopology(plan *models.ApmPlan) []interface{} {
+	var result = make([]interface{}, 0, len(plan.ClusterTopology))
+	for _, topology := range plan.ClusterTopology {
+		var m = make(map[string]interface{})
+
+		m["instance_configuration_id"] = topology.InstanceConfigurationID
+
+		m["memory_per_node"] = util.MemoryToState(*topology.Size.Value)
```
marclop, line 67: `Size` should also be checked for `nil`, and probably `Value` too since it might panic easily.

ec/ecdatasource/deploymentdatasource/state/flatteners_elasticsearch.go
```diff
@@ -0,0 +1,83 @@
...
+		var plan = res.Info.PlanInfo.Current.Plan
+		m["version"] = plan.Elasticsearch.Version
+
+		m["topology"] = flattenElasticsearchTopology(plan)
+
+		for k, v := range util.FlattenClusterEndpoint(res.Info.Metadata) {
+			m[k] = v
+		}
+
+		m["status"] = *res.Info.Status
```
marclop, line 50: Same for all the properties accessed that are pointers.

ec/ecdatasource/deploymentdatasource/state/flatteners_elasticsearch.go
```diff
@@ -0,0 +1,83 @@
...
+
+		m["node_count_per_zone"] = topology.NodeCountPerZone
+
+		m["node_type_data"] = *topology.NodeType.Data
+
+		m["node_type_ingest"] = *topology.NodeType.Ingest
+
+		m["node_type_master"] = *topology.NodeType.Master
+
+		m["node_type_ml"] = *topology.NodeType.Ml
```
marclop, line 77: Here too.

ec/ecdatasource/deploymentdatasource/state/flatteners_enterprise_search.go
```diff
@@ -0,0 +1,81 @@
...
+
+		m["elasticsearch_cluster_ref_id"] = *res.ElasticsearchClusterRefID
+
+		for k, v := range util.FlattenClusterEndpoint(res.Info.Metadata) {
+			m[k] = v
+		}
+
+		m["secret_session_key"] = plan.EnterpriseSearch.SystemSettings.SecretSessionKey
+
+		m["status"] = *res.Info.Status
```
marclop, line 52: Same

ec/ecdatasource/deploymentdatasource/state/flatteners_enterprise_search.go
```diff
@@ -0,0 +1,81 @@
...
+
+		m["node_type_appserver"] = *topology.NodeType.Appserver
+
+		m["node_type_connector"] = *topology.NodeType.Connector
+
+		m["node_type_worker"] = *topology.NodeType.Worker
+
+		m["zone_count"] = topology.ZoneCount
+
+		result = append(result, m)
```
marclop, line 77: Same

ec/ecdatasource/deploymentdatasource/state/flatteners_kibana.go
```diff
@@ -0,0 +1,73 @@
...
+
+		m["topology"] = flattenKibanaTopology(plan)
+
+		m["elasticsearch_cluster_ref_id"] = *res.ElasticsearchClusterRefID
+
+		for k, v := range util.FlattenClusterEndpoint(res.Info.Metadata) {
+			m[k] = v
+		}
+
+		m["status"] = *res.Info.Status
```
marclop, line 50: Same

ec/ecdatasource/deploymentdatasource/state/flatteners_kibana.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func flattenKibanaTopology(plan *models.KibanaClusterPlan) []interface{} {
+	var result = make([]interface{}, 0, len(plan.ClusterTopology))
+	for _, topology := range plan.ClusterTopology {
+		var m = make(map[string]interface{})
+
+		m["instance_configuration_id"] = topology.InstanceConfigurationID
+
+		m["memory_per_node"] = util.MemoryToState(*topology.Size.Value)
```
marclop, line 65: Same here

marclop (commented):
There's a very real possibility that deployments which have been stopped or have no current plan might case panics in several places of the codebase, left individual comments on it.

Also, From what I can see, we're only covering the use case where a user will know the full ID of the deployment not the shortened ID. I'm not sure we'd like to cover that case in this PR, but wanted to make sure that we're covering what we talked about with @Kushmaro

karencfv:
> There's a very real possibility that deployments which have been stopped or have no current plan might case panics

Ah yes! That's a really good point.

> Also, From what I can see, we're only covering the use case where a user will know the full ID of the deployment not the shortened ID. I'm not sure we'd like to cover that case in this PR, but wanted to make sure that we're covering what we talked about with @/ Kushmaro

Was going to write something about that in the PR's description, but forgot. Sorry! There are a few points on that:

- We won't be able to use a shortened version of IDs everywhere (e.g. `$ terraform import ec_deployment.search 320b7b540dfc967a7a649c18e2fce4ed` and probably others), so it might feel a bit strange for users to have an option in one place but not in others. I think I'd like to go for consistency.
- ecctl doesn't allow shortened IDs, and even though they are two different products, I think it makes sense to have some sort of consistency between ecctl and the provider. This way they have the feel of one cohesive set of tools.
- In order to allow for shortened IDs, we'd have to use the "search" endpoint instead of the "get" endpoint which is what is being used right now. I realise I could change the endpoint we use, but I believe we should use the endpoint that is intended for the action.
- From what I remember in that conversation, Omer mentioned something about how the full IDs will be displayed in the UI soon. I think we should develop the provider with that in mind.

Let me know what you think!

marclop:
Sounds good @karencfv!

karencfv:
```
=== RUN   TestAccDatasourceDeployment_basic
=== PAUSE TestAccDatasourceDeployment_basic
=== CONT  TestAccDatasourceDeployment_basic
--- PASS: TestAccDatasourceDeployment_basic (960.70s)
PASS
```

ec/ecdatasource/deploymentdatasource/datasource.go
```diff
@@ -0,0 +1,115 @@
...
+func modelToState(d *schema.ResourceData, res *models.DeploymentGetResponse) error {
+	if err := d.Set("name", res.Name); err != nil {
+		return err
+	}
+
+	if err := d.Set("healthy", res.Healthy); err != nil {
+		return err
+	}
+
+	es := res.Resources.Elasticsearch[0]
```
marclop, line 78: This direct access to the Elasticsearch slice position makes me nervous, but I have manually checked the status of shut down deployments. and they still have the ES element in there, so I guess there's no reason to guard it here.

ec/ecdatasource/deploymentdatasource/datasource.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if es.Region != nil {
+		if err := d.Set("region", *es.Region); err != nil {
+			return err
+		}
+	}
+
+	if es.Info != nil && es.Info.PlanInfo != nil &&
```
marclop, line 86: There's a `elasticsearchstate.IsCurrentPlanEmpty` which could be used here. Alternatively, the `ec_deployment` resource uses a private function (getDeploymentTemplateID), which we might want to move to `elasticsearchstate` and use in both places.

ec/ecdatasource/deploymentdatasource/state/flatteners_apm.go
```diff
@@ -0,0 +1,99 @@
...
+			if res.Info.PlanInfo != nil && res.Info.PlanInfo.Current != nil &&
+				res.Info.PlanInfo.Current.Plan != nil {
+				var plan = res.Info.PlanInfo.Current.Plan
+
+				if plan.Apm != nil {
+					m["version"] = plan.Apm.Version
+				}
+
+				if plan.Apm != nil && plan.Apm.SystemSettings != nil {
+					m["secret_token"] = plan.Apm.SystemSettings.SecretToken
```
marclop, line 63: Let's not populate this field. Access to it might be removed very soon from the user API.

ec/ecdatasource/deploymentdatasource/state/flatteners_elasticsearch.go
```diff
@@ -0,0 +1,112 @@
...
+
+			if res.Info.ClusterID != nil {
+				m["resource_id"] = *res.Info.ClusterID
+			}
+
+			if res.Info.Status != nil {
+				m["status"] = *res.Info.Status
+			}
+
+			if res.Info.PlanInfo != nil && res.Info.PlanInfo.Current != nil &&
```
marclop, line 50: We could use the same `elasticsearchstate.IsCurrentPlanEmpty` here

ec/ecdatasource/deploymentdatasource/state/flatteners_enterprise_search.go
```diff
@@ -0,0 +1,113 @@
...
+
+			if res.Info.ID != nil {
+				m["resource_id"] = *res.Info.ID
+			}
+
+			if res.Info.Status != nil {
+				m["status"] = *res.Info.Status
+			}
+
+			if res.Info.PlanInfo != nil && res.Info.PlanInfo.Current != nil &&
```
marclop, line 54: there's a `isCurrentPlanEmpty` in the `enterprisesearchstate` too.

ec/ecdatasource/deploymentdatasource/state/flatteners_apm.go
```diff
@@ -0,0 +1,99 @@
...
+
+			if res.Info.ID != nil {
+				m["resource_id"] = *res.Info.ID
+			}
+
+			if res.Info.Status != nil {
+				m["status"] = *res.Info.Status
+			}
+
+			if res.Info.PlanInfo != nil && res.Info.PlanInfo.Current != nil &&
```
marclop, line 54: there's a `isCurrentPlanEmpty` in the `apmstate` too.

ec/ecdatasource/deploymentdatasource/state/flatteners_enterprise_search.go
```diff
@@ -0,0 +1,113 @@
...
+				res.Info.PlanInfo.Current.Plan != nil {
+				var plan = res.Info.PlanInfo.Current.Plan
+
+				if plan.EnterpriseSearch != nil {
+					m["version"] = plan.EnterpriseSearch.Version
+				}
+
+				if plan.EnterpriseSearch != nil &&
+					plan.EnterpriseSearch.SystemSettings != nil {
+					m["secret_session_key"] = plan.EnterpriseSearch.SystemSettings.SecretSessionKey
```
marclop, line 64: Let's not map this key either, it's not accessible by users, it'll always be `""`

ec/ecdatasource/deploymentdatasource/state/flatteners_kibana.go
```diff
@@ -0,0 +1,94 @@
...
+
+			if res.Info.ClusterID != nil {
+				m["resource_id"] = *res.Info.ClusterID
+			}
+
+			if res.Info.Status != nil {
+				m["status"] = *res.Info.Status
+			}
+
+			if res.Info.PlanInfo != nil && res.Info.PlanInfo.Current != nil &&
```
marclop, line 54: there's a `isCurrentPlanEmpty` in the `kibanastate` too.

marclop (commented):
Nice work on documenting all the resource fields in Markdown! I left a few comments on functions we can make public and re-use

ec/ecdatasource/deploymentdatasource/datasource.go
karencfv: I had to set it this way unfortunately :( there's no other way to access the region

marclop: Yeah I know, it's all good, it's how the data is structured unfortunately 🙂

karencfv:
Thanks for the feedback @marclop !

marclop (approved):
Looks great! Thanks for addressing my comments 🚀

PR #64, opened 2020-09-09
Add outputs to "examples/deployment"

karmi:
This patch adds an example `outputs.tf` file to the examples,
which returns the credentials for connecting to the cluster.


marclop (changes requested):
It seems like the name `test_terraform` needs to be replaced with `example_minimal`

karmi:
Argh, sorry about that! Fixed!

marclop (approved):
LGTM

Issue #58, opened 2020-09-08
Run acceptance tests on every PR on Elastic CI


marclop:
## Overview

Currently, the integration tests aren't run on every pull request, while we publish what we run locally on a PR as a comment, it's not the most trustworthy method.

It would be preferable if the Acceptance tests were run on every PR via the Elastic CI.

marclop:
Acceptance tests are being run on every PR.

the `master` branch build doesn't seem to be triggered, I'll investigate some more

marclop:
I'm closing this one off since the acceptance tests are indeed running on every PR, I continue to investigate why it's still not working for `master`

PR #83, opened 2020-09-16
ci: Add jenkins jjbb job definitions

marclop:
## Description

Adds the JJBB definitions as pipeline jobs and a groovy Jenkinsfile to
run the acceptance tests in the `devops-ci` infra-owned Jenkins cluster.

There's two job definitions, one for master builds and another one for
pull requests, where the job will get automatically triggered if the PR
author is under the elastic GitHub org, otherwise a `jenkins test this`
is required for external contributors.

Additionally, adds a view which should group all the jobs containing
the `terraform-provider-ec` string.

## Related Issues

Part of #58 

## How Has This Been Tested?

😅


.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,45 @@
+---
+
+##### GLOBAL METADATA
+
+- meta:
+    cluster: devops-ci
```
alpar-t, line 6: Might be worth adding a view  and including the job into it  to make it easier to find the job when browsing the CI server.

.ci/jobs/elastic+terraform-provider-ec+master.yml
```diff
@@ -0,0 +1,31 @@
...
+        wipe-workspace: 'False'
+    triggers:
+    - github
+    axes:
+    - axis:
+        type: slave
+        name: label
+        values:
+        - docker&&virtual
+    builders:
```
alpar-t, line 18: Could this move up to the defaults ?

marclop:
WIP since I'll change the builders plugin in favor of the pipeline one.

.ci/pipelines/acceptance.Jenkinsfile
```diff
@@ -0,0 +1,38 @@
...
+    docker.image("${DOCKER_IMAGE}").inside("-u root:root -v ${pwd()}:${APP_PATH} -w ${APP_PATH}") {
+        try {
+            stage("Download dependencies") {
+                sh 'make vendor'
+            }
+            stage("Run acceptance tests") {
+                sh 'make testacc-ci'
+            }
+        } catch (Exception err) {
+            throw err
```
Aran-K, line 30: I don't think you don't actually need this since you're not doing anything with the exception, Groovy/Java don't require a `catch` block

.ci/jobs/elastic+terraform-provider-ec+pull-request.yml
```diff
@@ -0,0 +1,22 @@
...
+            allow-whitelist-orgs-as-admins: true
+            cancel-builds-on-update: true
+            status-context: acceptance
+    pipeline-scm:
+        scm:
+            - git:
+                branches:
+                    - $ghprbActualCommit
+                refspec: +refs/pull/*:refs/remotes/origin/pr/*
+                wipe-workspace: "False"
```
Aran-K, line 22: Does this work? I would have thought `false` was right but maybe jjb(b) is doing some type magic 🤷‍♂️

Aran-K (approved):
Just some minor things that caught my eye. 
I haven't set up this workflow (public repos/GHPRB etc) before but it looks fine to me 🙂

.ci/jobs/elastic+terraform-provider-ec+pull-request.yml
marclop: I'm not too sure about this one, it was a c/p from other projects settings. I don't think it even matters here for what we're doing, I'll remove it.

.ci/pipelines/acceptance.Jenkinsfile
marclop: from the top of my head, I remember this being necessary to be able to catch the errors, since they're thrown within Docker and otherwise the pipeline returned with success even when there was an error, but maybe that's from a while ago?

Aran-K: Hmm I'm not sure actually. That sounds like a very "Jenkins" thing that could well be true, it just stood out cause it reads funny. Not going to cause a problem if you leave it there though 🙂

marclop: yeah, I think I might just leave it there for now, and venture into removing it later since I'm sure this works 😂

.ci/jobs/elastic+terraform-provider-ec+pull-request.yml
Aran-K: It defaults to `true` and just means "don't bother cleaning the workspace since we use ephemeral workers", but there's also no harm in cleaning it, so it doesn't really matter either way I guess!

PR #86, opened 2020-09-18
ci: Fix broken acceptance build pipeline

marclop:
## Description

Fixes a few issues with the committed pipeline and build configurations.

First, The path used for the secret wasn't readable by the vault role
configured in the `devops-ci`, furthermore that key contained more
credentials in the secret path. Instead of using the initial path, a new
secret has been created in the Vault server with only the apikey
credentials, in a path accessible by the vault role.

Second, fixes the docker `golang` docker image to  use the tag  `1.15`.

Last, it changes the `acceptance.yml` github action to only run on the
push operation in the master branch. This is temporary and until the
master build in Jenkins is fixed (Currently not triggered).

## Related issues

Part of #58

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -18,4 +18,4 @@ clean: ## Delete credentials
 	@ rm -f .apikey
 
 .apikey:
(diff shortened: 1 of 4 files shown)


artemnikitin (approved):

Aran-K (approved):
LGTM!

PR #88, opened 2020-09-21
ec_deployments: New datasource

karencfv:
## Description
Adds a new data source which allows the terraform provider users to 
retrieve a list of deployments based on a specified query defined by
pre-set filters. 

Many more filters can be added in the future, but for now I'd like to 
keep it simple by having only the ones that are most likely to be used
(based on feedback from internal users).

I have added acceptance testing as part of the same test with the 
`ec_deployment` datasource. Given it's such a small test, it made 
sense to me to have it this way for test speed, and to avoid spinning up 
more instances than needed. I can separate it into it's own test if 
this seems like a bad idea.

## Related Issues
Closes: https://github.com/elastic/terraform-provider-ec/issues/8 
🎉 

## How Has This Been Tested?
```console
=== RUN   TestAccDatasourceDeployment_basic
=== PAUSE TestAccDatasourceDeployment_basic
=== CONT  TestAccDatasourceDeployment_basic
--- PASS: TestAccDatasourceDeployment_basic (413.37s)
PASS
ok  	github.com/elastic/terraform-provider-ec/ec/acc	414.917s
```

## Types of Changes

## Readiness Checklist


ec/ecdatasource/deploymentsdatasource/datasource.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return nil
+}
+
+func modelToState(d *schema.ResourceData, res *models.DeploymentsSearchResponse) error {
+	if err := d.Set("return_count", res.ReturnCount); err != nil {
+		return err
+	}
+
+	if res.ReturnCount != ec.Int32(0) {
```
marclop, line 79: This if seems a little it redundant, if len(`res.Deployments`) == 0, then the for look won't iterate over anything. We can just remove it and check the length of result to be greater than 0. before setting the "deployments".

ec/ecdatasource/deploymentsdatasource/schema.go
```diff
@@ -0,0 +1,118 @@
...
+			Optional: true,
+			MaxItems: 1,
+			Elem:     newResourceFilters(),
+		},
+	}
+}
+
+func newDeploymentList() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
```
marclop, line 76: Is it intentional to leave the model so bare? I thought we were going to map each of the found deployments with the same attributes as `ec_deployment`

ec/util/resource_kinds.go
```diff
@@ -0,0 +1,32 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package util
+
+const (
+	//Elasticsearch represents the elasticsearch resource kind
+	Elasticsearch = "elasticsearch"
```
marclop, line 22: We're duplicating this instead of using [`github.com/elastic/cloud-sdk-go/pkg/util`](https://pkg.go.dev/github.com/elastic/cloud-sdk-go/pkg/util#pkg-constants)

marclop (commented):
Good job on the filters! I left a question regarding if we should provide the models to be equal to what `ec_deployment` provides for each of the found deployments.

ec/ecdatasource/deploymentsdatasource/datasource.go
karencfv: I don't know why I did that 😄

ec/ecdatasource/deploymentsdatasource/schema.go
karencfv: Yes, this was intentional. That's one of the reasons I divided #8 into two datasources. This way one of the datasources (ec_deployment) would give you detailed information about a single resource and this datasource would only give you deployment IDs which you can later use for batch actions. Similar to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/autoscaling_groups

But I'm not attached to my solution, if you disagree with me or have some other use case in mind, happy to change it! :)

marclop: I guess that makes sense, I'm not opposed to it at all 👍

marclop (approved):

PR #98, opened 2020-09-28
ec_resource: Use deployment template as blueprint

### LINKED ISSUE #94: ec_deployment: Make instance_configuration_id optional
## Overview

Both the `deployment_template_id` and `topology.instance_configuration_id` are required, but in fact, we can automatically obtain the `topology.instance_configuration_id` by reading the deployment_template.

This is going to make getting started much easier and less confusing.

Good safeguards should be put in place which return an error when there's more than one topology element and instance_configuration_id is missing, but the API validation would already return that error.

marclop:
## Description

Changes the `topology` block within each of deployment resources from
`required` to `optional` by using the `deployment_template` template of
the `GET deployment templates` API call. Not only aligning with how the
Elastic Cloud Web UI works, but also benefiting from defaults, important
metadata and pre-configured settings set in the deployment template now
consumed, as a result a series of _side-effects_ have cascaded down to
the `ec_resource` schema itself, allowing users to create deployments
specifying much less, and making getting started much easier.

Both on `Create` and `Update`, the `GET /deployments/templates` endpoint
is called, filtering out the specified `deployment_template_id` field,
and reading the `.deployment_template` object which contains a blueprint
to create a deployment using this template.
The `deployment_template` contains all the resource kinds node_types,
curation settings, node attributes (for tiered data topologies), and
`instance_configuration_id`, all details that the user was burdened to
know before to be able to create a deployment. Since the Elasticsearch
resource is the most complex of resources, it's the one benefiting most
from this change.
Then, the template is used as a blueprint, and any attributes set by the
user will be overridden from the deployment template blueprint.

Also removing the default values for `memory_per_node` and `zone_count`.
This caused a little bit more of an impact on the code than expected,
thus all the handling of empty and / or unset topology settings and mem
size.

Since there's no reason to allow users to tweak the Elasticsearch and
Enterprise Search node types, they have been removed as settable, and 
are only computed for reference.

As part of this patch, the "region" field has been marked as `ForceNew`
since changing the region field will cause a deployment to be destroyed
and a new one created.

Additionally, two different bugs are fixed in this patch.

First, removing any resource from the Terraform definition didn't cause
the resource to be removed from the backend due to `prune_orphans` being
set to `false`.

Second, any resources which were removed, still appear on the plan with
"stopped" in the `status`, since no filtering was done on the flattening
functions, an infinite loop was triggered and they were attempted to be
removed again, and persisted in the state. Filtering any resources with
`status == "stopped"` fixes this.

Last, a series of new unit tests have been added which now test more
scenarios where users are some overrides on the topology to ensure that
the code behaves as it should.

### Minimal

```hcl
resource "ec_deployment" "example_minimal" {
  # Mandatory fields

  elasticsearch {}

  kibana {}

  apm {}

  enterprise_search {}
}
```

### Omitting the `instance_configuration_id`, but setting overrides

```hcl
# Create an Elastic Cloud deployment
resource "ec_deployment" "example_minimal" {
  # Mandatory fields

  elasticsearch {
    topology {
      memory_per_node = "1g"
    }
  }

  kibana {
    topology {
      memory_per_node = "2g"
    }
  }

  apm {
    topology {
      memory_per_node = "1g"
    }
  }

  enterprise_search {
    topology {
      memory_per_node = "4g"
    }
  }
}
```

## Related Issues

Closes #94 

## Motivation and Context

Try and reduce the Minimal configuration required

## How Has This Been Tested?

Thorougly manually and through the acceptance tests.

## Types of Changes


marclop:
@karencfv apologies for the big PR, but there wasn't a great way to split this. I have described to the best of my ability what all the changes are.

ec/ecresource/deploymentresource/apm_expanders.go
```diff
@@ -162,3 +160,28 @@ func expandApmConfig(raw interface{}) *models.ApmConfiguration {
...
+}
+
+func matchApmTopology(id string, topologies []*models.ApmTopologyElement) (*models.ApmTopologyElement, error) {
+	for _, t := range topologies {
+		if t.InstanceConfigurationID == id {
+			return t, nil
+		}
+	}
+	return nil, fmt.Errorf(
+		`apm topology: invalid instance_configuration_id: "%s" doesn't match any of the deployment template instance configurations`,
```
karencfv, line 184: Wondering if it's necessary to have this validation as the API validates this as well:
```console
 api error: cluster.invalid_instance_config: For template [I/O Optimized] the plan definition should contain instance configuration [aws.data.highio.i3]. Only [aws.coordinating.m5d] was supplied
```

ec/ecresource/deploymentresource/apm_expanders.go
```diff
@@ -162,3 +160,28 @@ func expandApmConfig(raw interface{}) *models.ApmConfiguration {
 
 	return nil
 }
+
+func defaultApmTopology(topology []*models.ApmTopologyElement) []*models.ApmTopologyElement {
+	for _, t := range topology {
+		if *t.Size.Value > defaultApmSize {
+			t.Size.Value = ec.Int32(defaultApmSize)
+		}
+		if t.ZoneCount > 1 {
```
karencfv, line 169: I don't really understand these checks? Why are we checking if size and zone count are larger than the defaults to set the defaults?

ec/ecresource/deploymentresource/delete.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 // Delete shuts down and deletes the remote deployment.
-func delete(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+func deleteFunc(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
karencfv, line 30: Why is this being changed? `delete` was fine?

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -201,3 +175,39 @@ func expandEsConfig(raw interface{}) *models.ElasticsearchConfiguration {
...
+}
+
+func matchEsTopology(id string, topologies []*models.ElasticsearchClusterTopologyElement) (*models.ElasticsearchClusterTopologyElement, error) {
+	for _, t := range topologies {
+		if t.InstanceConfigurationID == id {
+			return t, nil
+		}
+	}
+	return nil, fmt.Errorf(
+		`elasticsearch topology: invalid instance_configuration_id: "%s" doesn't match any of the deployment template instance configurations`,
```
karencfv, line 210: same comment as above

ec/ecresource/deploymentresource/enterprise_search_expanders.go
```diff
@@ -170,3 +153,28 @@ func expandEssConfig(raw interface{}) *models.EnterpriseSearchConfiguration {
...
+}
+
+func matchEssTopology(id string, topologies []*models.EnterpriseSearchTopologyElement) (*models.EnterpriseSearchTopologyElement, error) {
+	for _, t := range topologies {
+		if t.InstanceConfigurationID == id {
+			return t, nil
+		}
+	}
+	return nil, fmt.Errorf(
+		`enterprise_search topology: invalid instance_configuration_id: "%s" doesn't match any of the deployment template instance configurations`,
```
karencfv, line 177: same comment as above

ec/ecresource/deploymentresource/kibana_expanders.go
```diff
@@ -155,3 +153,28 @@ func expandKibanaConfig(raw interface{}) *models.KibanaConfiguration {
...
+}
+
+func matchKibanaTopology(id string, topologies []*models.KibanaClusterTopologyElement) (*models.KibanaClusterTopologyElement, error) {
+	for _, t := range topologies {
+		if t.InstanceConfigurationID == id {
+			return t, nil
+		}
+	}
+	return nil, fmt.Errorf(
+		`kibana topology: invalid instance_configuration_id: "%s" doesn't match any of the deployment template instance configurations`,
```
karencfv, line 177: same as above

ec/ecresource/deploymentresource/schema_elasticsearch.go
```diff
@@ -79,24 +81,27 @@ func elasticsearchTopologySchema() *schema.Schema {
...
-					Required:    true,
+					Description: `Optional Instance Configuration ID from the deployment template`,
+					Computed:    true,
+					Optional:    true,
 				},
 				"memory_per_node": {
 					Type:        schema.TypeString,
 					Description: `Optional amount of memory per node in the "<size in GB>g" notation`,
-					Default:     "4g",
+					Default:     util.MemoryToState(defaultElasticsearchSize),
```
karencfv, line 98: optional nit: This could be more readable if it was set as a variable beforehand. Something like `defaultElasticsearchSizeGB := util.MemoryToState(defaultElasticsearchSize)`. What do you think?

ec/ecresource/deploymentresource/schema_enteprise_search.go
```diff
@@ -66,18 +70,20 @@ func newEnterpriseSearchResource() *schema.Resource {
...
 				"instance_configuration_id": {
 					Type:     schema.TypeString,
-					Required: true,
+					Optional: true,
+					Computed: true,
 				},
 				"memory_per_node": {
 					Type:     schema.TypeString,
-					Default:  "2g",
+					Default:  util.MemoryToState(defaultEnterpriseSearchSize),
```
karencfv, line 86: same comment about setting it as a variable

ec/ecresource/deploymentresource/schema_enteprise_search.go
```diff
@@ -66,18 +70,20 @@ func newEnterpriseSearchResource() *schema.Resource {
 func enterpriseSearchTopologySchema() *schema.Schema {
 	return &schema.Schema{
 		Type:     schema.TypeList,
-		Required: true,
+		Optional: true,
+		Computed: true,
```
karencfv, line 74: Looks like these should have the `Description` fields set as well.

ec/ecresource/deploymentresource/schema_kibana.go
```diff
@@ -66,18 +70,20 @@ func newKibanaResource() *schema.Resource {
...
 				"instance_configuration_id": {
 					Type:     schema.TypeString,
-					Required: true,
+					Optional: true,
+					Computed: true,
 				},
 				"memory_per_node": {
 					Type:     schema.TypeString,
-					Default:  "1g",
+					Default:  util.MemoryToState(defaultKibanaSize),
```
karencfv, line 86: same comment about setting it as a variable

ec/ecresource/deploymentresource/schema_kibana.go
```diff
@@ -66,18 +70,20 @@ func newKibanaResource() *schema.Resource {
 func kibanaTopologySchema() *schema.Schema {
 	return &schema.Schema{
 		Type:     schema.TypeList,
-		Required: true,
+		Optional: true,
+		Computed: true,
```
karencfv, line 74: Same comment about using the `Description` field

docs/resources/ec_deployment.md
```diff
@@ -196,6 +180,10 @@ In addition to all arguments above, the following attributes are exported:
 * `elasticsearch.#.cloud_id` - The encoded Elasticsearch credentials to use in Beats or Logstash, [more information](https://www.elastic.co/guide/en/cloud/current/ec-cloud-id.html).
 * `elasticsearch.#.http_endpoint` - The Elasticsearch resource HTTP endpoint.
 * `elasticsearch.#.https_endpoint` - The Elasticsearch resource HTTPs endpoint.
+* `elasticsearch.#.topology.#.node_type_data` - (Optional) Node type (data) for the Elasticsearch Topology element.
+* `elasticsearch.#.topology.#.node_type_master` - (Optional) Node type (master) for the Elasticsearch Topology element.
+* `elasticsearch.#.topology.#.node_type_ingest` - (Optional) Node type (ingest) for the Elasticsearch Topology element.
+* `elasticsearch.#.topology.#.node_type_ml` - (Optional) Node type (machine learning) for the Elasticsearch Topology element.
```
karencfv, line 186: These are computed now

docs/resources/ec_deployment.md
```diff
@@ -211,6 +199,9 @@ In addition to all arguments above, the following attributes are exported:
 * `enterprise_search.#.region` - The Enterprise Search region.
 * `enterprise_search.#.http_endpoint` - The Enterprise Search resource HTTP endpoint.
 * `enterprise_search.#.https_endpoint` - The Enterprise Search resource HTTPs endpoint.
+* `enterprise_search.#.topology.#.node_type_appserver` - (Optional) Node type (Appserver) for the Enterprise Search Topology element.
+* `enterprise_search.#.topology.#.node_type_connector` - (Optional) Node type (Connector) for the Enterprise Search Topology element.
+* `enterprise_search.#.topology.#.node_type_worker` - (Optional) Node type (worker) for the Enterprise Search Topology element.
```
karencfv, line 204: These are computed now

ec/internal/util/testutil.go
```diff
@@ -0,0 +1,75 @@
...
+// template.
+func ApmResource(res *models.DeploymentTemplateInfoV2) *models.ApmPayload {
+	return res.DeploymentTemplate.Resources.Apm[0]
+}
+
+// EnterpriseSearchResource returns the EnterpriseSearchPayload from a deployment
+// template.
+func EnterpriseSearchResource(res *models.DeploymentTemplateInfoV2) *models.EnterpriseSearchPayload {
+	return res.DeploymentTemplate.Resources.EnterpriseSearch[0]
+}
```
karencfv, line 75: These are only used in the `deploymentresource` package. It might make sense to leave them in that package unless we see the need to use them somewhere else? I can't really think of any other use case for these functions outside of `deploymentresource` 🤔

karencfv (commented):
I really like the implementation 🏅 Thanks for working on this! I've left some comments but nothing too major.

karencfv:
hmmm... enterprise_search keeps failing on the acceptance tests for some reason 🤔

ec/ecresource/deploymentresource/delete.go
marclop: there is a function in expanders.go:172-176 which uses the built-in `delete` function to delete keys of a map:

```go
func unsetTopology(rawRes []interface{}, d *schema.ResourceData) {
	for _, r := range rawRes {
		delete(r.(map[string]interface{}), "topology")
	}
}
```

Since we need to use this and the private `delete` function was shadowing the built-in `delete` I renamed it here.

ec/ecresource/deploymentresource/apm_expanders.go
marclop: Yeah sorry, I'll add a comment on the code to solve that.

Basically, there's a few scenarios that I'm trying to cover here in the APM resource, at the moment, the deployment template is as follows: 

```javascript
{
    "ref_id": "apm-ref-id",
    "elasticsearch_cluster_ref_id": "es-ref-id",
    "region": "us-east-1",
    "plan": {
        "cluster_topology": [
            {
                "instance_configuration_id": "aws.apm.r5d",
                "size": {
                    "value": 512,
                    "resource": "memory"
                },
                "zone_count": 1
            }
        ],
        "apm": {}
    }
}
```

but this could change in an upcoming deployment template, to have a consistent behavior across the board, what I've done is:

-  if the size is > than what we say is the default in the `memory_per_node`, then an override with that default is set.
- if the zone_count is > 1 then, we apply the default 1.

Apm and Kibana have the same topology pre say, the only thing that changes is the `instance_configuration_id`, but the template for Enterprise Search is sets the `size.value: 0`, which then we need to check in that function as we do.

Elasticsearch is slightly different, where the ES deployment template contains all the available templates for each of the instance configurations enabling only a some of the blueprinted ones:

```javascript
{
    "ref_id": "es-ref-id",
    "region": "us-east-1",
    "plan": {
        "cluster_topology": [
            {
                "node_type": {
                    "master": false,
                    "data": false,
                    "ingest": true
                },
                "zone_count": 2,
                "instance_configuration_id": "aws.coordinating.m5d",
                "size": {
                    "value": 0,
                    "resource": "memory"
                }
            },
            {
                "node_type": {
                    "master": true,
                    "data": true,
                    "ingest": true
                },
                "zone_count": 2,
                "instance_configuration_id": "aws.data.highio.i3",
                "size": {
                    "value": 8192,
                    "resource": "memory"
                }
            },
            {
                "node_type": {
                    "master": true,
                    "data": false,
                    "ingest": false
                },
                "zone_count": 3,
                "instance_configuration_id": "aws.master.r5d",
                "size": {
                    "value": 0,
                    "resource": "memory"
                }
            },
            {
                "node_type": {
                    "master": false,
                    "data": false,
                    "ingest": false,
                    "ml": true
                },
                "zone_count": 1,
                "instance_configuration_id": "aws.ml.m5d",
                "size": {
                    "value": 0,
                    "resource": "memory"
                }
            }
        ],
        "elasticsearch": {}
    },
    "settings": {
        "dedicated_masters_threshold": 6
    }
}
```

In the case above ^, the instance configuration which is enabled by default is `"aws.data.highio.i3", which has a size of `8192` and `zone_count` set to `2`. This is not what we advertise as defaults in the terraform provider and thus, to provide a consistent experience, is why we check the ZoneCount and the size.

I've answered here all of the other comments, I'm going to mark the others as resolved and we can keep this conversation here.

ec/ecresource/deploymentresource/schema_enteprise_search.go
marclop: Probably not something to fix in this PR

ec/ecresource/deploymentresource/schema_elasticsearch.go
marclop: It wouldn't be my first choice, I left it as is to avoid too much indirection so we don't have an explosion of variables set in a place and then we have to go look what they are, if you feel very strongly about it I can change it

ec/ecresource/deploymentresource/apm_expanders.go
marclop: It's not strictly necessary, but it makes handling this case in the code so much easier, since otherwise a `nil` pointer to struct would be returned and that would need to get handled with exactly he same way, do you have any other way in mind?

karencfv: Ah! got it. Nah, it's fine :)

ec/ecresource/deploymentresource/apm_expanders.go
karencfv: Thanks for the explanation 🙇‍♀️ 

> In the case above ^, the instance configuration which is enabled by default is "aws.data.highio.i3", which has a size of 8192 and zone_count set to 2. This is not what we advertise as defaults in the terraform provider

Should we consider having those defaults on the provider as well? I'm thinking those defaults are on the template for a reason, maybe we shouldn't change them? 🤔

ec/ecresource/deploymentresource/delete.go
karencfv: Ah! of course! How about changing them all to `deleteResource`, `createResource`... (or `createDeployment`) etc? It makes me cringe a little to have `read`, `create`, `update`, `deleteFunc` 😄

ec/ecresource/deploymentresource/schema_elasticsearch.go
karencfv: nah, don't feel too strongly about it, we can leave it as is :)

ec/ecresource/deploymentresource/stopped_resource.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// isEssResourceStopped returns true if the resource is stopped.
+func isEssResourceStopped(res *models.EnterpriseSearchResourceInfo) bool {
+	return res == nil || res.Info == nil || res.Info.Status == nil ||
+		*res.Info.Status == "stopped"
+}
+
+// isKibanaResourceStopped returns true if the resource is stopped.
+func isKibanaResourceStopped(res *models.KibanaResourceInfo) bool {
```
karencfv, line 41: I didn't mention about moving these because I thought they'd be useful for https://github.com/elastic/terraform-provider-ec/issues/101 what do you think?

ec/ecresource/deploymentresource/testutil_func.go
```diff
@@ -0,0 +1,96 @@
...
+// essResource returns the EnterpriseSearchPayload from a deployment
+// template.
+func essResource(res *models.DeploymentTemplateInfoV2) *models.EnterpriseSearchPayload {
+	return res.DeploymentTemplate.Resources.EnterpriseSearch[0]
+}
+
+// flattenClusterEndpoint receives a ClusterMetadataInfo, parses the http and
+// https endpoints and returns a map with two keys: `http_endpoint` and
+// `https_endpoint`
+func flattenClusterEndpoint(metadata *models.ClusterMetadataInfo) map[string]interface{} {
```
karencfv, line 81: Are you thinking of removing the `util` package and copying the functions from there to every package that uses them? `util.FlattenClusterEndpoint` is used in several packages 🤔 
If that's the case maybe we should do that in a follow up PR? It feels a bit weird to have this function moved here now, have `util.FlattenClusterEndpoint` still exist, and still use `util.ParseTopologySize` in several places. What do you think?

ec/ecresource/deploymentresource/delete.go
marclop: haha okay the `<action>Resource` or `<action>Func` sound fine to me.

ec/ecresource/deploymentresource/apm_expanders.go
marclop: I was thinking about it... I think we should move to it, my only doubt are the size for the trials in ESS, I'm not sure they're allowed to spin 2x 8Gb, but I'll confirm this today.

ec/ecresource/deploymentresource/stopped_resource.go
marclop: yeah, but I don't see how keeping these in `util` would make any difference, restoring from snapshot will be the same resource, right?

ec/ecresource/deploymentresource/testutil_func.go
marclop: ignore, I mistakenly moved all the contents of the package here, and then realized what I did, undid it, but forgot to remove these.

ec/ecresource/deploymentresource/stopped_resource.go
karencfv: > restoring from snapshot will be the same resource, right?

🤦‍♀️

karencfv: 😂

ec/ecresource/deploymentresource/testutil_func.go
marclop: I am removing this function and its calls on the next commit.

Kushmaro:
awesome stuff @marclop !
few questions though on your writeup - 
>it does not rely on
the default memory size in the deployment_template, but instead a local
default is used for Elasticsearch (4Gb instead of 8Gb per node).

why do we use this local default?

>Since there's no reason to allow users to tweak the Elasticsearch and
Enterprise Search node types, they have been removed as settable, and
are only computed for reference.

could you please elaborate? does this align with all future plans we have for elasticsearch node types?

ec/ecresource/deploymentresource/create.go
```diff
@@ -28,14 +28,14 @@ import (
...
-func create(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+// createResource will createResource a new deployment from the specified settings.
+func createResource(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*api.API)
 	reqID := deploymentapi.RequestID(d.Get("request_id").(string))
 
 	req, err := createResourceToModel(d, client)
 	if err != nil {
-		diag.FromErr(err)
+		return diag.FromErr(err)
```
marclop, line 38: We weren't returning the error here 🤦

marclop (commented):
We weren't returning the error upon expanding the model... I've fixed it in this commit.

marclop:
> awesome stuff @marclop !
> few questions though on your writeup -
> 
> > it does not rely on
> > the default memory size in the deployment_template, but instead a local
> > default is used for Elasticsearch (4Gb instead of 8Gb per node).
> 
> why do we use this local default?

The description was outdated, I have now updated to reflect what the patch is changing, there's no more local defaults for `memory_per_node` or `zone_count`.

> 
> > Since there's no reason to allow users to tweak the Elasticsearch and
> > Enterprise Search node types, they have been removed as settable, and
> > are only computed for reference.
> 
> could you please elaborate? does this align with all future plans we have for elasticsearch node types?

We're just not allowing users to set those settings anymore since they are set by the deployment template on each of the instance configurations.
Now, we match the instance configuration specified (or not, then the default for that DT), and those instance configurations already have the "roles" defined, when I say node types I'm referring to `data`, `master`, `ingest`, `ml`. In any case I think that users are not allowed to specify different node_types than what's in the Instance Configuration anyway, so we might have saved users from a configuration which they could set but wasn't allowed anyway.

marclop:
@karencfv the `enterprise_search` instance seems to be bootlooping... 

```
Status: [Failed] Applying Actastic Migration: <UpdateUsersAndEngines version=20191229125356>: Error = Swiftype::ES::IndexAlreadyExistsException: index [.ent-search-actastic-users_v2/2ZPAnIE3R5es9poii8kGfA] already exists index '.ent-search-actastic-users_v2'
```

ec/ecresource/deploymentresource/create.go
karencfv: omg I didn't catch that either 🙊 glad you found it!

marclop:
@karencfv I fixed the integration tests on 060daf8

marclop:
Jenkins test this please.

karencfv (approved):
Thank you for fixing the test! 🙇‍♀️

ggsood:
Hi @marclop thanks for doing the change, this simplifies a lot of configuration, i was testing the new version, during which i found the below issue, for deployment template

```
resource "ec_deployment" "example_minimal" {
  # Optional name.
  name = "horizon-hot-warm-test"

  # Mandatory fields
  region                 = "aws-eu-west-2"
  deployment_template_id = "aws-hot-warm-v2"

  elasticsearch {}

  kibana {}
}
```
is getting created but if we update the configuration like
```
resource "ec_deployment" "example_minimal" {
  # Optional name.
  name = "horizon-hot-warm-test"

  # Mandatory fields
  region                 = "aws-eu-west-2"
  deployment_template_id = "aws-hot-warm-v2"

  elasticsearch {
    topology {
      instance_configuration_id = "aws.data.highio.i3"
      zone_count = 1
      memory_per_node = "1g"
    }
    topology {
      instance_configuration_id = "aws.data.highstorage.d2"
      zone_count = 1
      memory_per_node = "2g"
    }
  }

  kibana {}
}
```
the terraform is passing but the configuration is not updating at Elastic cloud and getting the error like
Elasticsearch - The requested cluster configuration change is not permitted. Value of field 'plan fields' cannot be changed.

Can u please also check updating the deployment works if it is provisioned from template.

marclop:
@ggsood that's strange, I'll try to reproduce locally and see what's going on.

ggsood:
H @marclop https://github.com/elastic/terraform-provider-ec/pull/108 this will help ?

marclop:
It will only mitigate part of the problem, the real issue is that we're overwriting the whole `Elasticsearch` object when the `config` part of the definition is expanded, I've described the issue in a comment: https://github.com/elastic/terraform-provider-ec/issues/111#issuecomment-704360511

PR #115, opened 2020-10-07
dependencies: Update terraform-plugin-sdk to v2.0.4

karencfv:
## Description
Updates the terraform-plugin-sdk dependency to v2.0.4

### DIFF
--- a/ec/acc/datasource_deployment_basic_test.go
+++ b/ec/acc/datasource_deployment_basic_test.go
@@ -40,7 +40,8 @@ func TestAccDatasourceDeployment_basic(t *testing.T) {
 		CheckDestroy:      testAccDeploymentDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: cfg,
+				Config:             cfg,
+				ExpectNonEmptyPlan: true,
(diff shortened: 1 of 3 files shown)


marclop (approved):
Looks good, but it worries me that the id changes when it's protected by that if clause https://github.com/elastic/terraform-provider-ec/blob/master/ec/ecdatasource/deploymentsdatasource/datasource.go#L62

karencfv:
@marclop I honestly don't remember why I put the `if` clause there. I can remove it, but I'll do it in a follow up PR as I feel like it falls out of the scope of this one

marclop:
oh, I think that the clause should be there, no need to re-generate the ID every single read?

karencfv:
Ah! right, got it. I misread what you had written. I hadn't noticed that until you mentioned it

Hmmmmm 🤔

karencfv:
Ah! I know why the ID changes. It's because the test has two checks. it tests `data.ec_deployment` which sets the ID to `d.SetId(deploymentID)` first, and `data.ec_deployments` which sets ID to `d.SetId(resource.UniqueId())` afterwards.

PR #117, opened 2020-10-07
version: Add "-dev" suffix to committed version

marclop:
## Description

Adds a `-dev` suffix  to the committed version Makefile, also modifying
the Go generator to enclose the variable in double quotes which was done
externally before.

## Motivation and Context

Avoid categorizing outgoing API calls to a released version when it might not.

## Types of Changes


karencfv (approved):
The tests passed! 😭

PR #118, opened 2020-10-07
config: Add verbose_ credentials and file setting

### LINKED ISSUE #110: Add verbose_credentials parameter to provider
## Overview

By default, the `request.log` of terraform should redact the APIKey and Authorization headers. To enable this and also enable un-redacting it and dumping the credentials in the file, a `verbose_credentials` attribute should be added to the provided to allow this.

## Possible Implementation

new `verbose_credentials` boolean which should map to the `EC_VERBOSE_CREDENTIALS` environment variable.

marclop:
## Description

This patch has two kinds of changes, user facing and a small refactoring
to allow the creation flow of `api.Config` to be unit tested.

User facing:

- By default, the Authorization header contents will now be redacted.
- Adds a new `verbose_credentials` boolean which allows users to dump
  the raw Authorization header in the log file.
- Adds a new `verbose_file` string which allows users to control the
  file where the request log is written to (defaults to `request.log`).
- Fixes a bug where the default timeout wasn't properly set to `1m`.
- Updates the `index.md` file.

Refactoring:

- Moves the provider schema to its own function to make easy to create
  *schema.ResourceData instances from it.
- `verboseSettings` wasn't returning an error on `os.Create` which just
  failed silently, this wasn't ideal since if users want debug to be
  enabled but the file path is incorrect, an error should be returned.
- Adds unit tests covering all code paths.

## Related Issues

Closes #110 

## How Has This Been Tested?

Unit & manually through the examples.

## Types of Changes

### DIFF
--- a/ec/provider.go
+++ b/ec/provider.go
@@ -36,15 +36,11 @@ const (
 	eceOnlyText      = "Available only when targeting ECE Installations or Elasticsearch Service Private"
 	saasRequiredText = "The only valid authentication mechanism for the Elasticsearch Service"
 
-	endpointDesc = "Endpoint where the terraform provider will point to. Defaults to \"%s\"."
-	insecureDesc = "Allow the provider to skip TLS validation on its outgoing HTTP calls."
-	timeoutDesc  = "Timeout used for individual HTTP calls. Defaults to \"1m\"."
-	verboseDesc  = "When set, a \"request.log\" file will be written with all outgoing HTTP requests. Defaults to \"false\"."
-)
-
-var (
-	// DefaultEndpoint is the default provider endpoint.
-	DefaultEndpoint = api.ESSEndpoint
+	endpointDesc     = "Endpoint where the terraform provider will point to. Defaults to \"%s\"."
+	insecureDesc     = "Allow the provider to skip TLS validation on its outgoing HTTP calls."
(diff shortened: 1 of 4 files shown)


ec/provider.go
```diff
@@ -136,3 +68,86 @@ func Provider() *schema.Provider {
 		},
 	}
 }
+
+func newSchema() map[string]*schema.Schema {
```
karencfv, line 72: I'm not sure about extracting the schema into it's own function? It doesn't have nested lists or sets 🤔 To me it feels more readable to have it like it was. that way when you're looking at the `Provider` struct, you don't have to go look for the `newSchema` function.

marclop: Yeah I understand that, but I'm not sure how we'd unit test this then? If you see in the unit tests, they're using `newSchema()` to make sure the schema is equal to the one in the provider, I'd rather use `newSchema` than a sloppy c/p?

marclop: https://github.com/elastic/terraform-provider-ec/pull/118/files#diff-c75bbf3696fdeba7a90420204358c3a4R108

karencfv: Fair point 🤔

karencfv (approved):

marclop:
Thanks!

Issue #11, opened 2020-05-26
ec_deployment: Make acceptance test fields dynamic


marclop:
## Description

Currently the acceptance tests for `ec_deployment` uses hardcoded  fields like:

* version (`7.6.2`)
* region (`us-east-1`)
* deployment_template_id (`aws-io-optimized`) and their instance configurations, Elasticsearch, Kibana and APM.

It would be best to make these parameters more dynamic depending on which region the tests are run in, and if run against an ECE environment should also work.

karencfv:
Putting this on hold until https://github.com/elastic/terraform-provider-ec/pull/98 is merged as there are a few changes that impact the work being done here :)

marclop:
Also reading the issue again, I think we shouldn't make the version too dynamic, what we're doing works pretty well, but we could use the `ec_stack` data source to get the stack version.

PR #121, opened 2020-10-11
acceptance: Make test fields dynamic

karencfv:
## Description
All `region`, `deployment_template_id` and `instance_configuration_id` fields
in the acceptance tests are now dynamic. Region fields are set through a 
new environment variable `EC_REGION`, or set to a default if this variable
is not set.

Only ESS regions are supported. ECE environments are not supported as 
the deployment template names and configurations will vary depending 
on the ECE installation.

Additionally, the files within the `testdata/` directory have been modified
to use default values rather specifying topology when it was not necessary 
to test this way. This reduced complexity in our test templates.

## Related Issues
Closes: https://github.com/elastic/terraform-provider-ec/issues/11

## How Has This Been Tested?
By running the tests with different cloud providers and regions

## Types of Changes

- [X] Refactoring (improves code quality but has no user-facing effect)

## Readiness Checklist

### DIFF
--- a/build/Makefile.test
+++ b/build/Makefile.test
@@ -27,7 +27,7 @@ tests: unit
 ## Runs the Terraform acceptance tests. Use TEST_NAME, TESTARGS and TEST_COUNT to control execution
 testacc:
 	@ echo "-> Running terraform acceptance tests..."
-	@ TF_ACC=1 go test $(TEST_ACC) -v -count $(TEST_COUNT) -parallel 20 $(TESTARGS) -timeout 120m -run $(TEST_NAME) -tags=acceptance
+	@ TF_ACC=1 go test $(TEST_ACC) -v -count $(TEST_COUNT) -parallel 20 $(TESTARGS) -timeout 120m -run $(TEST_NAME)
 
 .PHONY: sweep
 ## Destroys any dangling infrastructure created by the acceptance tests (terraform_acc_ prefix).
--- a/build/Makefile.test
+++ b/build/Makefile.test
@@ -27,7 +27,7 @@ tests: unit
 ## Runs the Terraform acceptance tests. Use TEST_NAME, TESTARGS and TEST_COUNT to control execution
 testacc:
 	@ echo "-> Running terraform acceptance tests..."
-	@ TF_ACC=1 go test $(TEST_ACC) -v -count $(TEST_COUNT) -parallel 20 $(TESTARGS) -timeout 120m -run $(TEST_NAME)
+	@ TF_ACC=1 go test $(TEST_ACC) -v -count $(TEST_COUNT) -parallel 20 $(TESTARGS) -timeout 120m -run $(TEST_NAME) -tags=acceptance
 
 .PHONY: sweep
 ## Destroys any dangling infrastructure created by the acceptance tests (terraform_acc_ prefix).
--- a/ec/acc/deployment_basic_defaults_test.go
+++ b/ec/acc/deployment_basic_defaults_test.go
@@ -18,6 +18,8 @@
 package acc
 
 import (
+	"fmt"
+	"io/ioutil"
 	"testing"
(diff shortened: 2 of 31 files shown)


ggsood:
Hi @karencfv thanks for the changes, would do you suggest having Region field as part of provider and all other resources then refer from there.

karencfv:
Hi @ggsood! In this case I am only introducing a new environment variable for the purposes of testing. This way it's not necessary to have a specific region hard coded into our acceptance tests.

I am not sure if I'd like to have this environment variable set for anything other than tests, as many people will have deployments in several different regions 🤔

ggsood:
yes @karencfv my apologies i was thinking it extends more than the test files, very sorry :(

karencfv:
No worries at all @ggsood! :)

build/Makefile.test
```diff
@@ -27,7 +27,7 @@ tests: unit
 ## Runs the Terraform acceptance tests. Use TEST_NAME, TESTARGS and TEST_COUNT to control execution
 testacc:
 	@ echo "-> Running terraform acceptance tests..."
-	@ TF_ACC=1 go test $(TEST_ACC) -v -count $(TEST_COUNT) -parallel 20 $(TESTARGS) -timeout 120m -run $(TEST_NAME)
+	@ TF_ACC=1 go test $(TEST_ACC) -v -count $(TEST_COUNT) -parallel 20 $(TESTARGS) -timeout 120m -run $(TEST_NAME) -tags=acceptance
```
marclop, line 30: I think we can skip running the tests that open the fixtures file by renaming them to `fixture...`

ec/acc/deployment_basic_defaults_test.go
```diff
@@ -164,3 +168,15 @@ func TestAccDeployment_basic_defaults(t *testing.T) {
 		},
 	})
 }
+
+func testAccDeploymentResourceBasicDefaults(t *testing.T, fileName, name, region, depTpl string) string {
```
marclop, line 172: These functions are called since their name starts with `test...`, If we change them to be called `fixture...` no need to add the `acceptance` tag to a ll the files and Make target.

marclop (commented):
Approach looks good! Left some minor comments

ec/acc/deployment_config_test.go
```diff
@@ -15,9 +15,153 @@
...
+		Region: getRegion(),
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return res.Stacks[0].Version, nil
+}
+
+func setDefaultTemplate(region, template string) string {
```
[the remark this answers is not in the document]
karencfv: I initially had `ece-region` return the default template as well, but we we use more than one template for the tests :( 
The `datasource` tests would still work I guess, but the hot/warm tests would cause trouble 🤔 Also, we cannot guarantee the deployment template Ids in people's ECE installations :/

ec/acc/deployment_basic_defaults_test.go
karencfv: 🤦‍♀️ 😄

karencfv:
```
--- FAIL: TestAccDeployment_basic (0.00s)
Error:     deployment_basic_test.go:159: authwriter: 1 error occurred:
        	* one of apikey or username and password must be specified
```

Looks like we do need to skip building these tests after all

marclop: yeah I you're right, since the tests themselves start with `Test....

marclop (approved):
Looks great!

ec/acc/deployment_config_test.go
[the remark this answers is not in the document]
marclop: alright, we can leave the ECE use case out for now anyway.

PR #125, opened 2020-10-13
ec_deployment: Support legacy plans through import

### LINKED ISSUE #10: ec_deployment: Test against legacy plans
## Description

At this point, no testing has been performed against legacy deployments with legacy plans, some testing should be performed to ensure it works interchangeably with legacy and current plans.

marclop:
## Description

Adds all of the necessary pieces too allow "legacy" (pre-ECE v2 and very
old ESS deployments which haven't changed in a looong time) deployments
to be managed by terraform through `terraform import`.

The first problem is that the `"region"` and `"version"` weren't being
set to the plan on the read, which caused the whole deployment to be
tainted when imported (region changes from `""` => `"<value>"`, causing
the whole deployment to be re-created (ForceNew field in effect).

Second, `deploymentapi.Get` needed `ConvertLegacyPlans` set to `true`.

Last, the deployment topology elements were being ignored since the size
is `0` because `memory_per_node` and `node_count_per_zone` are used in
the legacy plan definitions. These changes only needed to be made in the
`elasticsearch` and `kibana` resource kinds since they're the only ones
that existed pre-DNT.

## Related Issues

Closes #10 

## How Has This Been Tested?

Manually with a very old deployment I own, which I then added as the unit test fixtures.

## Types of Changes


karencfv (commented):
Thinking about what you had said about how many people actually use legacy plans, I also wonder whether they're worth supporting. Supporting them may cause us more problems down the line. The provider is a new product, and we already decided to not support Appsearch, should we really have support for something so old? 

Looking at the code, I realise legacy plans do not have a boolean to tell us they are legacy or any other easy way to determine this, but I'm not sure I like having to rely on certain "unique" features of legacy plans to determine whether a plan is legacy or not. I get the feeling that this could cause us some headaches in the future.

What do you think @marclop @Kushmaro ?

marclop:
That sounds very reasonable and would be my preference as well. 

I'll close this Pull Request, and open a new one with some of the enhancements (mostly tests) and the region / version field I had made.

PR #126, opened 2020-10-14
ec_deployment: Fix broken user_settings_json field

### LINKED ISSUE #124: user_settings_json field doesn't work as a string.
## Expected Behavior

Users can set a both `user_settings_json` and `user_settings_overrides_json`.

## Current Behavior

Currently, we're passing settings `cfg.UserSettingsJSON` and `cfg.UserSettingsOverridesJSON` both type `interface{}` as a string, which doesn't work when that field is actually set in the plan, it wasn't failing until now because there's no test case actually covering that field.

It's an interface to be able to accept a raw JSON, in which case it should be a `map[string]interface{}` which is set to the field.

## Possible Solution

Use the wonderful `json` library.

marclop:
## Description

Fixes `user_settings_json` and `user_settings_overrides_json` content
parsing in the expanders and flatteners where the value of the field was
being set as is instead of being handled correctly through the `json`
library which Marshals / Unmarshals the contents of the field.

Additionally, a little bit of a cleanup to provide consistent behavior
for the config expanders has been done too, where the destination struct
is passed as a pointer instead of being returned, and now an error is
returned (should any `json` encoding / decoding occur).

Furthermore, the `region` and `version` field have been added to the
flatteners causing the import acceptance tests to not have to ignore the
fields anymore.

Last, a few test fixtures and cases for different templates have been
added as well, including a case in the acceptance tests where the apm
resource has a `user_settings_json` set, since json and yaml settings
cannot be included at the same time.

## Related Issues

Closes #124 

## How Has This Been Tested?

Manual, acceptance and unit test.

## Types of Changes


karencfv (approved):
🏅 Looks great!

PR #127, opened 2020-10-14
docs: Add note on importing legacy deployments

marclop:
## Description

Adds a warning note on the `ec_deployment` resource about importing pre-
Slider deployments.

## Related Issues

Part of #10
Action item from #125 

## Motivation and Context

Doesn't make them work, but at least there's a note about it

## Types of Changes

- [x] Documentation

### DIFF
--- a/docs/resources/ec_deployment.md
+++ b/docs/resources/ec_deployment.md
@@ -225,7 +225,7 @@ In addition to all arguments above, the following attributes are exported:
 
 ## Import
 
-~> **Note on legacy (pre-slider) deployments** If your deployment was created prior to the addition of Sliders in both ECE or ESS, and has not been migrated over to use Sliders, importing it is not supported and will result in erroneous behavior.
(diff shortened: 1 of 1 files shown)


docs/resources/ec_deployment.md
```diff
@@ -225,6 +225,8 @@ In addition to all arguments above, the following attributes are exported:
 
 ## Import
 
+~> **Note on legacy (pre-slider) deployments** If your deployment was created prior to the addition of Sliders in both ECE or ESS, and has not been migrated over to use Sliders, importing it is not supported and will result in erroneous behavior.
```
karencfv, line 228:
```suggestion
~> **Note on legacy (pre-slider) deployments** If your deployment was created prior to the addition of sliders in both ECE or ESS, and has not been migrated to use sliders, importing it is not supported and will result in erroneous behavior.
```

karencfv (approved):

PR #128, opened 2020-10-14
config: Reduce timeout to 40s

marclop:
## Description

Reduce the timeout to `40s` from `1m` since any potential client network
instability needn't take 1m to resolve and is likely that a retry might
make the request succeed when retried.

Still `40s` is a reasonable timeout time for most requests.

## Motivation and Context

Retries should take care of timed out requests and `1m` timeout seems excessive for the the timeouts that are most likely to be cause by client network issues.

## How Has This Been Tested?

running `make testacc`

## Types of Changes

- [X] Refactoring (improves code quality but has no user-facing effect)

### DIFF
--- a/ec/provider_config_test.go
+++ b/ec/provider_config_test.go
@@ -26,7 +26,6 @@ import (
 	"path/filepath"
 	"syscall"
 	"testing"
-	"time"
 
 	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/auth"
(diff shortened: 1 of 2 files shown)


ec/provider_config_test.go
```diff
@@ -103,6 +102,19 @@ func Test_verboseSettings(t *testing.T) {
 }
 
 func Test_newAPIConfig(t *testing.T) {
+	// This is necessary to avoid any EC_API_KEY which might be set to cause
+	// test flakyness.
+	if k := os.Getenv("EC_API_KEY"); k != "" {
```
karencfv, line 107: Is this common? I've not really come across this issue?

Maybe we can extract it into it's own function? I think it'll make it look a bit cleaner, what do you think?

marclop: Yeah it's pretty common if you're running the acceptance tests locally and have an `EC_API_KEY` set, then the tests will fail. Maybe you're not running the acceptance tests very often these days 🙊, haha just kidding!

marclop: I'll extract it into its own function with a `t.Helper() set`

karencfv: acceptance tests 😖

karencfv (approved):
Thanks for addressing my comments!

marclop:
I am merging this regardless, the failing tests are unrelated.

Issue #129, opened 2020-10-15
Terraform Provider not supporting few ES Templates


ggsood:
The current version of terraform provider does not support creation of certain ES templates like CCS clusters.
The CCS clusters doesn't have an APM, Enterprise Search and App Search Resource.
The code currently checks them in the template API Response,  code checks need to be handled in this case.

## Readiness Checklist
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
The CCS cluster should get created as per the template.

## Current Behavior
Getting the following error when running the terraform resource

```hcl
resource "ec_deployment" "defaults" {
  name                   = "ccs-cluster"
  region                 = "us-east-1"
  version                = "7.9.2"
  deployment_template_id = "aws-cross-cluster-search-v2"

  elasticsearch {
    topology {
      size = "1g"
    }
  }

  kibana {
    topology {
      size = "1g"
    }
  }
}
```

## Possible Solution
Add nil checks for APM and Enterprise Resources in deployment template.

## Steps to Reproduce
Run the terraform code provided above with the latest version of code.

## Context
Unable to create CCS cluster

## Your Environment

* Version used: master
* Environment name and version (e.g. Go 1.9): 1.13
* Server type and version: Mac
* Operating System and version: MacOS

marclop:
You forgot to include the returned error, only posted the `hcl` definition, I'd be keen to know what error was thrown.

ggsood:
Apologies for that @marclop , please find the error stack below
```
panic: runtime error: index out of range [0] with length 0
2020-10-14T17:44:15.070+0100 [DEBUG] plugin.terraform-provider-ec_v1.0.1: 
2020-10-14T17:44:15.070+0100 [DEBUG] plugin.terraform-provider-ec_v1.0.1: goroutine 23 [running]:
2020-10-14T17:44:15.070+0100 [DEBUG] plugin.terraform-provider-ec_v1.0.1: github.com/elastic/terraform-provider-ec/ec/ecresource/deploymentresource.apmResource(...)
2020-10-14T17:44:15.070+0100 [DEBUG] plugin.terraform-provider-ec_v1.0.1:       github.com/elastic/terraform-provider-ec/ec/ecresource/deploymentresource/testutil_func.go:68
[...]

PR #130, opened 2020-10-15
ec_deployment: ignore empty resources

ggsood:
Fix code to handle empty resources.

## Description
In CCS cluster template, APM, AppSearch and Enterprise Search are not included, the code doesn't have a check to handle these resource in case they are not present, adding nil checks to handle such scenarios

## Related Issues
Addresses Issue https://github.com/elastic/terraform-provider-ec/issues/129

## Motivation and Context
To Create CCS cluster

## How Has This Been Tested?
- [x] Acceptance Tests
- [x] Unit tests

## Types of Changes

## Readiness Checklist


ec/ecresource/deploymentresource/expanders.go
```diff
@@ -60,22 +60,25 @@ func createResourceToModel(d *schema.ResourceData, client *api.API) (*models.Dep
...
-		return nil, err
+	if res.DeploymentTemplate.Resources.Apm != nil {
+		apmRes, err := expandApmResources(
+			d.Get("apm").([]interface{}), apmResource(res),
+		)
+		if err != nil {
+			return nil, err
+		}
+		result.Resources.Apm = append(result.Resources.Apm, apmRes...)
+	}
```
marclop, line 71: I'd rather guard a single line with an `if` clause than all the lines above.

```suggestion
	apmRes, err := expandApmResources(
		d.Get("apm").([]interface{}), apmResource(res),
	)
	if err != nil {
		return nil, err
	}
	if len(apmRes) > 0 {
		result.Resources.Apm = append(result.Resources.Apm, apmRes...)
	}
```

ec/ecresource/deploymentresource/expanders.go
```diff
@@ -60,22 +60,25 @@ func createResourceToModel(d *schema.ResourceData, client *api.API) (*models.Dep
...
+
+	if res.DeploymentTemplate.Resources.EnterpriseSearch != nil {
+		enterpriseSearchRes, err := expandEssResources(
+			d.Get("enterprise_search").([]interface{}), essResource(res),
+		)
+		if err != nil {
+			return nil, err
+		}
+		result.Resources.EnterpriseSearch = append(result.Resources.EnterpriseSearch, enterpriseSearchRes...)
 	}
```
marclop, line 81: Same here

```suggestion
	enterpriseSearchRes, err := expandEssResources(
		d.Get("enterprise_search").([]interface{}), essResource(res),
	)
	if err != nil {
		return nil, err
	}
	if len(enterpriseSearchRes) > 0 {
		result.Resources.EnterpriseSearch = append(result.Resources.EnterpriseSearch, enterpriseSearchRes...)
	}
```

ec/acc/deployment_basic_defaults_test.go
```diff
@@ -165,6 +167,24 @@ func TestAccDeployment_basic_defaults(t *testing.T) {
 					resource.TestCheckResourceAttr(resName, "enterprise_search.#", "0"),
 				),
 			},
+			{
+				// Change the Elasticsearch resource deployment template to
+				// CCS, use defaults.
+				Config: ccsCfg,
```
marclop, line 173: Actually this test will fail, because it's going from a Hot / Warm, where `node_attributes` are set, to using a CCS deployment_template which does not have `node_attributes`.

This test should be a standalone test and not a test which requires on a previous state.

marclop (commented):
Thank you for the contribution.

The fix is a valid fix, but it doesn't only fix a CCS template which you found to not be working as expected.

Changing the title to something like: `ec_deployment: Ignore empty resources` would be more accurate.

ggsood:
Thanks @marclop have changed the title.

ec/acc/deployment_basic_defaults_test.go
ggsood: thanks @marclop will move this to a separate test case.

ec/ecresource/deploymentresource/expanders.go
ggsood: this might now work because the error is thrown from inside `apmResource` function

ec/acc/deployment_basic_defaults_test.go
ggsood: have updated issue #129 with the error message

ec/ecresource/deploymentresource/expanders.go
ggsood: Have updated #129  with the error message

ec/ecresource/deploymentresource/expanders.go
ggsood: same as the above message.

ec/acc/deployment_basic_defaults_test.go
ggsood: Hi @marclop have created new test case file for it.

marclop:
@ggsood I totally misread your PR. There's somewhat of an issue on the approach that you took:

If the deployment template doesn't include the resource kind (Apm, or Enterprise Search), then the resource will be ignored altogether, meaning that even if the user has defined an `apm {}` or `enterprise_search {}` block, those resources will be ignored, even when the deployment allows that resource kind to be set.

A better way to fix this bug would be, modifying the `esResource`, `kibanaResource`, `apmResource` and `essResource`, checking the `len()` of the resource kind in question and returning `nil` when `len() == 0`

Then the nil case should be handled at each of the resource expanding functions (`expandEsResources`, `expandKibanaResources`, `expandApmResources` and `expandEssResources`.

It's actually a bug I introduced when working on #94, and looking at it requires a few more changes throughout the codebase, I'd rather tackle it myself and fix what I broke.

It should add a few more test cases and scenarios where this bug is reproducible, and then introduce the fix in the code.

ggsood:
No worries @marclop, i can close this PR for now then, please do suggest if i can help with anything.

PR #132, opened 2020-10-15
terraform: Bump to 0.13.x and strip install suffix

marclop:
## Description

This PR updates both the terraform version to the latest `0.13.x` and
also removes the `-<prefix>` that the VERSION variable might contain.

## Motivation and Context

testing no longer cumbersome with `0.13.x`.

## Types of Changes

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Install terraform
       uses: hashicorp/setup-terraform@v1
       with:
-        terraform_version: "0.12.29"
+        terraform_version: "0.13.x"
 
     - name: Cache Go Modules
       uses: actions/cache@v1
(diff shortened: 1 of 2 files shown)


marclop:
I'll debug this locally with `act`

Kushmaro:
Question @marclop ,
Would this mean we'll not test at all with 0.12.x?

I think we should make the provider both 12 and 13 compatible if we can.
v11 had some annoying changes, i'm not sure if the diff between 12->13 was that big though.

thoughts?

marclop:
This change only affects the `validate-examples` target.

For acceptance testing we rely on what's packaged in the SDK version, which I think uses `terraform v0.1.13.x` but it gets updated regularly.

I too think we might want to do matrix testing at some point, but I'm unsure atm how to go about this with the acceptance testing. I can open a ticket so we can investigate.

Kushmaro:
SGTM @marclop , thanks

karencfv (approved):
🙇‍♀️

Issue #60, opened 2020-09-08
Add monitoring via log delivery APIs


marclop:
## Overview

Add monitoring object which uses the "log delivery" mechanism.

## Possible Implementation

Use the new Log delivery APIs available after MS46 is deployed.

karencfv:
Blocked until MS46 is released

marclop:
@Kushmaro how similar do should we make this to our UI?

<img width="1043" alt="image" src="https://user-images.githubusercontent.com/7286993/95852716-dd8ba180-0d5c-11eb-8c7a-90074e5e4189.png">

```hcl
resource "ec_deployment" "lnm" {
  ...
  observability {
    deployment_id = "some target deployment id"
    logs          = true
    metrics       = true
  }
}
```

The API accepts 2 different destinations (maybe you want to send logs to a deployment and metrics to another). So technically we could send to different destinations, that would look like:

```hcl
resource "ec_deployment" "lnm" {
  ...
  observability {
    logs    = "some target deployment id"
    metrics = "some target deployment id"
  }
}
```

I'll also let @karencfv chime in since she's assigned this issue to herself, maybe she's got some good ideas here too.

ggsood:
I think we also need to update elastic cloud sdk version for it

marclop:
> I think we also need to update elastic cloud sdk version for it

It's already up to date, no need to update it.

Kushmaro:
@marclop the API accepting two locations is for flexibility / later API expansion if needed.
we shouldn't need to this at this time.
I'm ++ on
observability {
    deployment_id = "some target deployment id"
    logs          = true
    metrics       = true
 }

can we default for both being true?

marclop:
Yes, we can set both to true by default, as long as we document it properly I don't see any problem with that.

karencfv:
Ha! Good thing you asked @marclop, I had already thought of using

```tf
resource "ec_deployment" "lnm" {
  ...
  observability {
    logs    = "some target deployment id"
    metrics = "some target deployment id"
  }
}
```

Although...

> the API accepting two locations is for flexibility  / later API expansion. We shouldn't need to this at this time.

The difference from the UI and the provider, is that with the former one day you'll find a new button that gives you the option to have 2 destinations which doesn't disrupt anyone, but with the provider this would be somewhat annoying for our users to change if we want to give them the option later on. When they update to a newer version their configurations will break and they'll have to change them. 

[...]

PR #138, opened 2020-10-19
ec_deployment: Add observability settings

karencfv:
## Description
Adds a new `observability` block to the `ec_deployment` resource.

With these settings users will be able to  shipping logs and monitoring 
metrics to a dedicated deployment where they are stored separately. 

There are three attributes in this new block:

* `observability.#.deployment_id` - Destination deployment ID for the shipped logs and monitoring metrics.
* `observability.#.logs` - Enables or disables shipping logs (defaults to true).
* `observability.#.metrics` - Enables or disables shipping metrics (defaults to true).

### Example usage

```hcl
resource "ec_deployment" "observability" {

  elasticsearch {}

  observability {
    deployment_id = ec_deployment.basic.id
  }
}
```

## Related Issues
Closes https://github.com/elastic/terraform-provider-ec/issues/60

## How Has This Been Tested?
manually, unit and acceptance tests

## Types of Changes

## Readiness Checklist


karencfv:
I'm not sure why the unit tests are failing here but passing locally 🤔 Looking into it

ec/acc/deployment_observability_test.go
```diff
@@ -0,0 +1,90 @@
...
+// This test case takes that on a hot/warm "ec_deployment", a select number of
+// topology settings can be changed without affecting the underlying Deployment
+// Template.
+func TestAccDeployment_observability(t *testing.T) {
+	resName := "ec_deployment.observability"
+	secondResName := "ec_deployment.basic"
+	randomName := prefix + acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	startCfg := "testdata/deployment_observability_1.tf"
+	updateCfg := "testdata/deployment_observability_2.tf"
+	secondUpdateCfg := "testdata/deployment_observability_3.tf"
```
marclop, line 40: We need to add a 4th scenario where the destination deployment is removed and so is the observability block, then we should assert that `"observability.#", "0"`

ec/acc/deployment_observability_test.go
```diff
@@ -0,0 +1,90 @@
...
+				Config: secondCfg,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttrPair(resName, "observability.0.deployment_id", secondResName, "id"),
+					resource.TestCheckResourceAttr(resName, "observability.0.metrics", "false"),
+					resource.TestCheckResourceAttr(resName, "observability.0.logs", "true"),
+				),
+			},
+			{
+				Config:             thirdCfg,
+				ExpectNonEmptyPlan: true,
```
marclop, line 69: Why is this necessary? When this field is enabled, it makes me suspect that we're sweeping something under the rug 😄

ec/ecresource/deploymentresource/flatteners.go
```diff
@@ -83,6 +83,12 @@ func modelToState(d *schema.ResourceData, res *models.DeploymentGetResponse) err
 				return err
 			}
 		}
+
+		if observability := flattenObservability(res.Settings); observability != nil {
```
marclop, line 87: I think checking the length is slightly more solid

```suggestion
		if observability := flattenObservability(res.Settings); len(observability) > 0 {
```

ec/ecresource/deploymentresource/observability.go
```diff
@@ -0,0 +1,97 @@
...
+	// If either of them is not nil the deployment ID will be filled.
+	if settings.Observability.Metrics != nil {
+		m["deployment_id"] = settings.Observability.Metrics.Destination.DeploymentID
+		m["metrics"] = true
+	}
+
+	if settings.Observability.Logging != nil {
+		m["deployment_id"] = settings.Observability.Logging.Destination.DeploymentID
+		m["logs"] = true
+	}
```
marclop, line 46: heh, this looks funny due to the difference between API and TF provider

ec/ecresource/deploymentresource/observability.go
```diff
@@ -0,0 +1,97 @@
...
+		params := deploymentapi.PopulateRefIDParams{
+			Kind:         util.Elasticsearch,
+			API:          client,
+			DeploymentID: depID.(string),
+			RefID:        ec.String(""),
+		}
+
+		if err := deploymentapi.PopulateRefID(params); err != nil {
+			return nil, err
+		}
```
marclop, line 75: lots of magic to be able to populate the ref ID here, but we have no choice from what I've seen in the codebase, unless we choose to pass the refID as a parameter to not incur the cost of the API call.

How about we change the `expandObservability` signature to allow a string, and if that is empty, we still perform this call? I think that'd be a little bit better and most cases we wouldn't perform this API call.

Additionally, I think we should be adding some more context to the error, saying `observability ref_id auto discovery: ` maybe.

ec/ecresource/deploymentresource/observability_test.go
```diff
@@ -0,0 +1,243 @@
...
+			want: &models.DeploymentObservabilitySettings{
+				Logging: &models.DeploymentLoggingSettings{
+					Destination: &models.AbsoluteRefID{
+						DeploymentID: &mock.ValidClusterID,
+						RefID:        ec.String("main-elasticsearch"),
+					},
+				},
+				//	Metrics: &models.DeploymentMetricsSettings{
+				//	Destination: &models.AbsoluteRefID{},
+				//	},
```
marclop, line 198: I think we can remove this

ec/ecresource/deploymentresource/observability_test.go
```diff
@@ -0,0 +1,243 @@
...
+				v: []interface{}{map[string]interface{}{
+					"deployment_id": mock.ValidClusterID,
+					"metrics":       true,
+					"logs":          false,
+				}},
+			},
+			want: &models.DeploymentObservabilitySettings{
+				//	Logging: &models.DeploymentLoggingSettings{
+				//	Destination: &models.AbsoluteRefID{},
+				//	},
```
marclop, line 227: Same here

marclop (commented):
approach looks good, left a couple of minor comments

marclop:
I forgot to add this in the review, but we should also remove all the `monitor_settings` references and handling:

* https://github.com/elastic/terraform-provider-ec/blob/master/ec/ecresource/deploymentresource/elasticsearch_expanders.go#L84-L100
* https://github.com/elastic/terraform-provider-ec/blob/master/ec/ecresource/deploymentresource/elasticsearch_flatteners.go#L175

ec/acc/deployment_observability_test.go
karencfv: Same 😄 I honestly don't know what's going on here. Initially I didn't have that field set to true and it passed locally but failed in the PR (as you can see in the commit history). I was really confused and then realised I was on terraform `0.13.2` and jenkins was running the latest version which is `0.13.4` and alas, when I updated my tf version to 0.13.4 my local tests had the same result as the tests on our CI and I had to change this field.

I'm not really sure what else to do?

ec/ecresource/deploymentresource/observability.go
karencfv: 🙈 it looks awful, but couldn't think of anything better. Got any ideas?

ec/ecresource/deploymentresource/observability.go
karencfv: Yeah 🤔 I was thinking of making it easier for our users given that ref_id doesn't seem to be used much by people. But I like your idea to include `ref_id` as an optional argument in the schema.

ec/ecresource/deploymentresource/observability_test.go
karencfv: 🤦‍♀️

ec/ecresource/deploymentresource/observability_test.go
karencfv: moar 🤦‍♀️

ec/acc/deployment_observability_test.go
marclop: I think this might be a side effect of the `monitoring_settings`: https://github.com/elastic/terraform-provider-ec/pull/138#issuecomment-711723134

ec/ecresource/deploymentresource/observability.go
marclop: I think this is perfectly fine, could "hide" it in a private function, but I don't think it's worth hiding the cost.

ec/acc/deployment_observability_test.go
karencfv: 🤞

ec/ecresource/deploymentresource/observability.go
```diff
@@ -0,0 +1,106 @@
...
+		m["metrics"] = true
+	}
+
+	if settings.Observability.Logging != nil {
+		m["deployment_id"] = settings.Observability.Logging.Destination.DeploymentID
+		m["ref_id"] = settings.Observability.Logging.Destination.RefID
+		m["logs"] = true
+	}
+
+	return []interface{}{m}
```
marclop, line 52: The problem with `- observability {}` block is here, above this block a check to see if `len(m) > 0` before returning `[]interface{}{m}` is required.

marclop (commented):
Mostly looks good, we still have to remove that `ExpectNonEmptyPlan` boolean in the test, but after applying the patch I'm pasting below, the acceptance tests pass just fine for me:

```diff
diff --git a/ec/ecresource/deploymentresource/flatteners.go b/ec/ecresource/deploymentresource/flatteners.go
index 6c438da..9a677b1 100644
--- a/ec/ecresource/deploymentresource/flatteners.go
+++ b/ec/ecresource/deploymentresource/flatteners.go
@@ -84,7 +84,7 @@ func modelToState(d *schema.ResourceData, res *models.DeploymentGetResponse) err
 			}
 		}
 
-		if observability := flattenObservability(res.Settings); observability != nil {
+		if observability := flattenObservability(res.Settings); len(observability) > 0 {
 			if err := d.Set("observability", observability); err != nil {
 				return err
 			}
diff --git a/ec/ecresource/deploymentresource/observability.go b/ec/ecresource/deploymentresource/observability.go
index e0c0baf..0b8c750 100644
--- a/ec/ecresource/deploymentresource/observability.go
+++ b/ec/ecresource/deploymentresource/observability.go
@@ -34,7 +34,6 @@ func flattenObservability(settings *models.DeploymentSettings) []interface{} {
 	}
 
 	var m = make(map[string]interface{})
-
 	// We are only accepting a single deployment ID and refID for both logs and metrics.
 	// If either of them is not nil the deployment ID and refID will be filled.
 	if settings.Observability.Metrics != nil {
@@ -49,6 +48,10 @@ func flattenObservability(settings *models.DeploymentSettings) []interface{} {
 		m["logs"] = true
 	}
 
+	if len(m) == 0 {
+		return nil
+	}
+
 	return []interface{}{m}
 }
 
@@ -57,8 +60,7 @@ func expandObservability(raw []interface{}, client *api.API) (*models.Deployment
 		return nil, nil
 	}
 
-	req := models.DeploymentObservabilitySettings{}
-
+	var req models.DeploymentObservabilitySettings
 	for _, rawObs := range raw {
 		var obs = rawObs.(map[string]interface{})
 
diff --git a/ec/ecresource/deploymentresource/observability_test.go b/ec/ecresource/deploymentresource/observability_test.go
index 5b32f74..4815f14 100644
--- a/ec/ecresource/deploymentresource/observability_test.go
+++ b/ec/ecresource/deploymentresource/observability_test.go
@@ -44,6 +44,10 @@ func TestFlattenObservability(t *testing.T) {
 			name: "flattens no observability settings when empty",
 			args: args{settings: &models.DeploymentSettings{}},
 		},
+		{
+			name: "flattens no observability settings when empty",
+			args: args{settings: &models.DeploymentSettings{Observability: &models.DeploymentObservabilitySettings{}}},
+		},
 		{
 			name: "flattens observability settings",
 			args: args{settings: &models.DeploymentSettings{

```

karencfv:
Ha! thanks for finding that bug 🎉

marclop (approved):
Looks good, let's merge. Looks like I created some conflicts for you merging my PR first, sorry about that!

karencfv:
Merging as error is unrelated to this PR, and tests had already passed before merging master branch

PR #141, opened 2020-10-19
acceptance: Remove acceptance build tag

marclop:
## Description

Removes the acceptance build tag and instead adds a new function which
calls `t.Skip()` when a function with a `*testing.T` as the first arg
requires external HTTP access, in which case it is skipped, to trigger
this behavior any tests which fall into that category should call
`requiresAPIConn` at the beginning of the function to ensure that we're
not incurring any errors when calling `newAPI()`.

## Related Issues

`acceptance` build tag introduced in #121

## Motivation and Context

This was bothering me since writing tests was becoming quite cumbersome
due to the go build avoidance on the `+build acceptance` tag.

## Types of Changes

### DIFF
--- a/ec/acc/acc_prereq.go
+++ b/ec/acc/acc_prereq.go
@@ -15,8 +15,6 @@
 // specific language governing permissions and limitations
 // under the License.
 
-// +build acceptance
-
 package acc
 
 import (
@@ -65,5 +63,11 @@ func testAccPreCheck(t *testing.T) {
 
 	if apikey != "" && (username != "" || password != "") {
 		t.Fatal("Only one of API Key or Username / Password can be specified to execute acceptance tests")
+	}
+}
+
+func requiresAPIConn(t *testing.T) {
+	if os.Getenv("TF_ACC") != "1" {
+		t.Skip()
 	}
 }
--- a/ec/acc/acc_prereq.go
+++ b/ec/acc/acc_prereq.go
@@ -66,6 +66,9 @@ func testAccPreCheck(t *testing.T) {
 	}
 }
 
+// requiresAPIConn should be called in functions which would be executed by the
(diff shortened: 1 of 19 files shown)


ec/acc/deployment_basic_test.go
```diff
@@ -63,24 +61,18 @@ func TestAccDeployment_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resName, "traffic_filter.#", "0"),
 				),
 			},
-			// Ensure that no diff is generated.
-			{Config: cfg, PlanOnly: true},
```
karencfv, line 67: Not opposed to this change, but wondering why we are removing these?

ec/acc/acc_prereq.go
```diff
@@ -67,3 +65,9 @@ func testAccPreCheck(t *testing.T) {
 		t.Fatal("Only one of API Key or Username / Password can be specified to execute acceptance tests")
 	}
 }
+
+func requiresAPIConn(t *testing.T) {
```
karencfv, line 69: Let's put a comment about what this function does so future contributors know when/how to use it :)

karencfv (commented):
Ooooh fancy 💅

ec/acc/deployment_basic_test.go
marclop: They're not necessary. I forgot to why I did it in the provider.

I added this at the beginning when writing the provider tests because I thought that TF didn't refresh the state after apply, which it does automatically now by design. So this step is already "happening" after every apply and if there's a plan diff, an error is returned, which has happened to me before.

karencfv (approved):

marclop:
Failures have nothing to do with this change, I'm merging anyway.

PR #151, opened 2020-10-22
acceptance: Set parallel tests to 4

marclop:
## Description

This change sets the maximum number of tests allowed to run to default
to `4` unless `TEST_ACC_PARALLEL` is set to a different value by the
user.

This avoids resource starvation on test running machines.

Additionally, updates the `cloud-sdk-go` dependency to include the last
change on the client timeouts.

## How Has This Been Tested?

running `make acceptance` while connected to a slow network.

## Types of Changes


marclop:
This patch should also alleviate the CI load and issues we're seeing.

karencfv (approved):
🎉

Issue #166, opened 2020-11-04
Curation Settings missing in Elastic Deployment Resource


ggsood:
Curation settings is missing in the Elastic Deployment Resource.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
Optional Curation settings should be present when configuring an ES Deployment.
This is important when template type is hot-warm or hot-warm-v2

## Current Behavior
Curation settings are currently missing in the ES resource, absence of this setting causes the following issues
1. Converts Hot Warm deployments to curation based setting - which will soon be deprecated
2. Cannot enable migration from the Cloud UI, the option does not work 
3. Disables ILM option in the Kibana
4. Does not adhere to the API JSON generated from UI, example

```json
POST /api/v1/deployments?validate_only=false
{
  "resources": {
    "elasticsearch": [
      {
        "region": "aws-eu-west-2",
        **"settings": {
          "curation": {
            "specs": []
          },**
          "dedicated_masters_threshold": 6
        },
        "plan": {
          "cluster_topology": [
            {
              "elasticsearch": {
                "node_attributes": {
                  "data": "hot"
                },
                "enabled_built_in_plugins": []
              },
              "node_type": {
                "data": true,
                "master": true,
                "ingest": true
              },
              "instance_configuration_id": "aws.data.highio.i3",
              "zone_count": 1,
              "size": {
                "value": 1024,
                "resource": "memory"
              }
            },
            {
              "elasticsearch": {
                "node_attributes": {
                  "data": "warm"
                },
                "enabled_built_in_plugins": []
              },
              "node_type": {
                "data": true,
                "master": false,
                "ingest": true
              },
              "instance_configuration_id": "aws.data.highstorage.d2",
              "zone_count": 1,
              "size": {
                "value": 2048,
                "resource": "memory"
              }
            }
          ],
          "elasticsearch": {
            "version": "7.9.3"
          },
          "deployment_template": {
            "id": "aws-hot-warm-v2"
          }
        },
        "ref_id": "main-elasticsearch"
      }
    ],
    "enterprise_search": [],
    "kibana": [
      {
        "elasticsearch_cluster_ref_id": "main-elasticsearch",
        "region": "aws-eu-west-2",
        "plan": {
          "cluster_topology": [
            {
              "instance_configuration_id": "aws.kibana.r5d",
[...]

PR #167, opened 2020-11-04
ec_deployment: Support Hot Warm ILM deployments

ggsood:
The PR is a feature switch for enabling and disabling curation settings for elasticsearch deployments, this is useful when deployment template category is of type hot warm.

Also as curation is deprecated in the future releases having this safety check enables current deployments to easily transition to ILM till the Templates are updated.

This PR will address issue https://github.com/elastic/terraform-provider-ec/issues/166

## How Has This Been Tested?
 - [x] Unit tests and linting is passed
 - [x] Create and update deployments on EC

## Readiness Checklist
 - [ ] I have updated the documentation accordingly, will do once change is reviewed

### DIFF
--- a/ec/ecresource/deploymentresource/elasticsearch_expanders.go
+++ b/ec/ecresource/deploymentresource/elasticsearch_expanders.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/elastic/cloud-sdk-go/pkg/models"
 	"github.com/elastic/cloud-sdk-go/pkg/util/ec"
@@ -64,6 +65,11 @@ func expandEsResource(raw interface{}, res *models.ElasticsearchPayload) (*model
 			res.Region = ec.String(r)
 		}
 	}
+
+	// This is necessary as the default Elasticsearch payload
+	// for hot warm deployments still enables curation based
+	// settings which will soon be deprecated.
+	res = setEmptyCuration(*res.Plan.DeploymentTemplate.ID, res)
 
 	if rt, ok := es["topology"]; ok && len(rt.([]interface{})) > 0 {
 		topology, err := expandEsTopology(rt, res.Plan.ClusterTopology)
@@ -228,4 +234,12 @@ func esResource(res *models.DeploymentTemplateInfoV2) *models.ElasticsearchPaylo
 		}
 	}
 	return res.DeploymentTemplate.Resources.Elasticsearch[0]
+}
+
+func setEmptyCuration(template string, esPayload *models.ElasticsearchPayload) *models.ElasticsearchPayload {
+	if strings.Contains(template, "hot-warm") {
+		esPayload.Plan.Elasticsearch.Curation = nil
+		esPayload.Settings.Curation = nil
+	}
+	return esPayload
 }
--- a/ec/ecresource/deploymentresource/elasticsearch_expanders_test.go
+++ b/ec/ecresource/deploymentresource/elasticsearch_expanders_test.go
@@ -252,30 +252,12 @@ func Test_expandEsResource(t *testing.T) {
 					RefID:  ec.String("main-elasticsearch"),
 					Settings: &models.ElasticsearchClusterSettings{
 						DedicatedMastersThreshold: 6,
-						Curation: &models.ClusterCurationSettings{
-							Specs: []*models.ClusterCurationSpec{
-								{
-									IndexPattern:           ec.String("logstash-*"),
-									TriggerIntervalSeconds: ec.Int32(86400),
-								},
-								{
-									IndexPattern:           ec.String("filebeat-*"),
-									TriggerIntervalSeconds: ec.Int32(86400),
-								},
-								{
-									IndexPattern:           ec.String("metricbeat-*"),
-									TriggerIntervalSeconds: ec.Int32(86400),
-								},
-							},
-						},
+						Curation:                  nil,
 					},
 					Plan: &models.ElasticsearchClusterPlan{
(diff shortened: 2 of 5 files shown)


karencfv (changes requested):
First of all thank you @ggsood for finding this bug and opening a PR to fix it 🎉 👏 

I'd like to change the approach to the solution though.

We want to keep the Schema as simple as possible so adding a new argument to the top level to account for a case that is only really necessary for Hot Warm deployments, would kind of move us toward the opposite direction.

Second, as this is a new product that hasn't been released yet, we also do not want to support things that are on their way to being deprecated (as you can tell, on the UI you can't create new hot-warm deployments with curation based settings). 

An approach that aligns with these two statements would be as follows:

- Handle setting empty curation settings within the Elasticsearch expanders.
- Modify the documentation on importing to include a warning note explaining how only ILM Hot Warm deployments are supported.

Like this:
```console
$ git diff
diff --git a/docs/resources/ec_deployment.md b/docs/resources/ec_deployment.md
index 283255e..e133900 100644
--- a/docs/resources/ec_deployment.md
+++ b/docs/resources/ec_deployment.md
@@ -304,6 +304,8 @@ In addition to all arguments above, the following attributes are exported:
 
 ~> **Note on legacy (pre-slider) deployments** If your deployment was created prior to the addition of sliders in both ECE or ESS, and has not been migrated to use sliders, importing it is not supported and will result in erroneous behavior.
 
+~> **Note on pre ILM Hot Warm deployments** If your deployment uses Index Curation, you will have to migrate to Index Lifecycle Management (ILM) before importing. Index Curation will be deprecated soon and is not supported in the provider.
+
 Deployments can be imported using the `id`, e.g.
 
diff --git a/ec/ecresource/deploymentresource/elasticsearch_expanders.go b/ec/ecresource/deploymentresource/elasticsearch_expanders.go
index 6e938d0..7fc5006 100644
--- a/ec/ecresource/deploymentresource/elasticsearch_expanders.go
+++ b/ec/ecresource/deploymentresource/elasticsearch_expanders.go
@@ -21,6 +21,7 @@ import (
        "encoding/json"
        "fmt"
        "reflect"
+       "strings"
 
        "github.com/elastic/cloud-sdk-go/pkg/models"
        "github.com/elastic/cloud-sdk-go/pkg/util/ec"
@@ -65,6 +66,11 @@ func expandEsResource(raw interface{}, res *models.ElasticsearchPayload) (*model
                }
        }
 
+       // This is necessary as the default Elasticsearch payload
+       // for hot warm deployments still enables curation based
+       // settings which will soon be deprecated.
+       res = setEmptyCuration(*res.Plan.DeploymentTemplate.ID, res)
+
        if rt, ok := es["topology"]; ok && len(rt.([]interface{})) > 0 {
                topology, err := expandEsTopology(rt, res.Plan.ClusterTopology)
                if err != nil {
@@ -229,3 +235,12 @@ func esResource(res *models.DeploymentTemplateInfoV2) *models.ElasticsearchPaylo
        }
        return res.DeploymentTemplate.Resources.Elasticsearch[0]
 }
+
+func setEmptyCuration(template string, esPayload *models.ElasticsearchPayload) *models.ElasticsearchPayload {
+       if strings.Contains(template, "hot-warm") {
+               esPayload.Plan.Elasticsearch.Curation = nil
+               esPayload.Settings.Curation = nil
+       }
+
+       return esPayload
+}
diff --git a/ec/ecresource/deploymentresource/expanders.go b/ec/ecresource/deploymentresource/expanders.go
index 08ad33f..b724d8a 100644
--- a/ec/ecresource/deploymentresource/expanders.go
+++ b/ec/ecresource/deploymentresource/expanders.go
@@ -93,11 +93,6 @@ func createResourceToModel(d *schema.ResourceData, client *api.API) (*models.Dep
        }
        result.Settings.Observability = observability
 
-       if d.Get("disable_curation").(bool) && len(result.Resources.Elasticsearch) > 0 {
-               result.Resources.Elasticsearch[0].Plan.Elasticsearch.Curation = nil
-               result.Resources.Elasticsearch[0].Settings.Curation = nil
-       }
-
        return &result, nil
 }
 
@@ -184,10 +179,6 @@ func updateResourceToModel(d *schema.ResourceData, client *api.API) (*models.Dep
                result.Settings.Observability = &models.DeploymentObservabilitySettings{}
        }
 
-       if d.Get("disable_curation").(bool) && len(result.Resources.Elasticsearch) > 0 {
-               result.Resources.Elasticsearch[0].Plan.Elasticsearch.Curation = nil
-               result.Resources.Elasticsearch[0].Settings.Curation = nil
-       }
        return &result, nil
 }
 
diff --git a/ec/ecresource/deploymentresource/schema.go b/ec/ecresource/deploymentresource/schema.go
index c7c797c..d19a475 100644
--- a/ec/ecresource/deploymentresource/schema.go
+++ b/ec/ecresource/deploymentresource/schema.go
@@ -59,12 +59,6 @@ func newSchema() map[string]*schema.Schema {
                        Description: "Optional request_id to set on the create operation, only use when previous create attempts return with an error and a request_id is returned as part of the error",
                        Optional:    true,
                },
-               "disable_curation": {
-                       Type:        schema.TypeBool,
-                       Description: "Optional disable curation settings in case of hot-warm to enable ILM settings",
-                       Optional:    true,
-                       Default:     false,
-               },
 
                // Computed ES Creds
                "elasticsearch_username": {
```

ggsood:
Thanks @karencfv for the review, have corrected the code and unit tests, please can you check.

karencfv:
Jenkins test this please

karencfv (approved):
Thank you @ggsood ! 🙇‍♀️

Issue #154, opened 2020-10-27
Being able to create a deployment from snapshot


Kushmaro:
## Overview
Enabling customers to create new deployments from existing deployment snapshots.

## Context
It's a handy feature in the UI, would make sense to have available in the provider as well.

PR #170, opened 2020-11-06
ec_deployment: Support snapshot source restore

karencfv:
## Description
Adds support to restore data from a snapshot of another deployment
with the addition of a new `snapshot_source` block. This block has the 
following attributes:

* `elasticsearch.#.snapshot_source.#.source_cluster_id` - ID of the Elasticsearch 
cluster that will be used as the source of the snapshot.
* `elasticsearch.#.snapshot_source.#.snapshot_name` - Name of the snapshot to 
restore. Use `__latest_success__` to get the most recent successful snapshot.

### Example implementation
```
resource "ec_deployment" "example_snapshot_source" {

  elasticsearch {
    snapshot_source {
      source_cluster_id = "4f9c0cda4fe64b319a641ba6ee2689f7"
      snapshot_name = "__latest_success__"
    }
  }
}
```

## Related Issues
Closes: https://github.com/elastic/terraform-provider-ec/issues/154

## How Has This Been Tested?
Manually by applying the plan with and without the `snapshot_name`
attribute. 

_Note: This block cannot have an acceptance test, as it takes 30 minutes
from de creation of a deployment for the first snapshot to be created._

## Types of Changes

## Readiness Checklist


ggsood:
Hi @karencfv thanks for fixing this issue, a stupid question from my side, how will the es deployment work with respect to terraform once the resource is created, mainly will it always restore the snapshot if a configuration change is done on the resource.

karencfv:
That's a very good question @ggsood. Yes, it will always restore the snapshot if a configuration change is made to the resource. 

It's probably a good idea for me to add a warning text about that on the documentation 🤔

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -174,6 +183,22 @@ func expandEsConfig(raw interface{}, esCfg *models.ElasticsearchConfiguration) e
 	return nil
 }
 
+func expandSnapshotSource(raw interface{}, restore *models.RestoreSnapshotConfiguration) error {
```
marclop, line 186: This function never returns any error, maybe a good idea to just remove the `error` return from the signature?

ec/ecresource/deploymentresource/elasticsearch_flatteners.go
```diff
@@ -71,6 +71,10 @@ func flattenEsResources(in []*models.ElasticsearchResourceInfo, name string, rem
 			m["config"] = c
 		}
 
+		if s := flattenSnapshotSource(plan); len(s) > 0 {
```
marclop, line 74: Since the restore snapshot is supposed to be a one-off (since it lives in the transient block part of the plan) operation I think that it makes sense to omit the flattening so that any time that the `elasticsearch.#.snapshot_source` is set, the user gets a diff with additions on that part.

This would also require documenting the behavior but it would have less disruptive consequences than flattening and re-applying that state at a later point in time.

marclop (commented):
I think we should omit the flattening of that part of the plan since it's meant to be a transient state and we shouldn't be reapplying that generally, see my comment on the flattening code.

ec/ecresource/deploymentresource/elasticsearch_flatteners.go
karencfv: Yeah, I was on the fence about adding the flattening behaviour 🤔. Given it already confused someone, I think you're right and it's best to remove it.

docs/resources/ec_deployment.md
```diff
@@ -168,13 +168,15 @@ The optional `elasticsearch.remote_cluster` block can be set multiple times to r
...
-##### Snapshot Source
+##### Snapshot source
 
-The optional `elasticsearch.snapshot_source` block restores data from a snapshot of another deployment. It supports the following:
+The optional `elasticsearch.snapshot_source` block, which restores data from a snapshot of another deployment, supports the following arguments:
 
 * `source_cluster_id` (Required) ID of the Elasticsearch cluster that will be used as the source of the snapshot.
 * `snapshot_name` (Optional) Name of the snapshot to restore. Use `__latest_success__` to get the most recent successful snapshot (Defaults to `__latest_success__`).
 
+~> **Note on behavior** The `snapshot_source` block represents transient settings on the deployment plan. The values of this block will not be flattened, and therefore not saved in the Terraform state. This means that whenever the `snapshot_source` block is set when running `terraform apply`, a snapshot will be restored.
```
marclop, line 178:
```suggestion
~> **Note on behavior** The `snapshot_source` block will not be saved in the Terraform state due to its transient nature. This means that whenever the `snapshot_source` block is set, a snapshot will **always** be restored, unless removed before running `terraform apply`.
```

docs/resources/ec_deployment.md
```diff
@@ -167,6 +168,15 @@ The optional `elasticsearch.remote_cluster` block can be set multiple times to r
 * `ref_id` (Optional) Remote elasticsearch `ref_id`, it is best left to the default value (Defaults to `main-elasticsearch`).
 * `ignore_unavailable` (Optional) If true, skip the cluster during search when disconnected (Defaults to `false`).
 
+##### Snapshot source
+
+The optional `elasticsearch.snapshot_source` block, which restores data from a snapshot of another deployment, supports the following arguments:
+
+* `source_cluster_id` (Required) ID of the Elasticsearch cluster that will be used as the source of the snapshot.
```
marclop, line 175: I missed this in the previous review, is the ID the deployment ID? or the Elasticsearch Cluster ID? It seems that it's the Elasticsearch Cluster ID, not the deployment ID + `ref_id`. This can cause a lot of confusion for users if they're referencing the deployment ID instead of the ES cluster ID.

Maybe we can add a note that states that this ID is not to be confused with the deployment ID and that it's the Elasticsearch Resource ID.

Maybe it even makes sense to rename this field to `source_elasticsearch_resource_id`? it's extremely long, but very verbose in where the ID comes from.

karencfv: Yeah 🤔  I can see how that can be confusing

marclop (approved):
Looks good!

I'm still a bit on the fence on having a pre-created deployment available in the dev account, but that kind of defeats the purpose of repeatable acceptance tests.

Something we could do is having a `fixtures` target which can be used to create a deployment with terraform. In any case, this would be an improvement on this PR as it would add a bunch of new concepts (fixtures, and the need for them when running some tests).

karencfv:
> I'm still a bit on the fence on having a pre-created deployment available in the dev account, but that kind of defeats the purpose of repeatable acceptance tests.

I had a thought about this when I was trying to think on how to add acc tests, but then running locally or on different regions would be problem as the deployment to take the snapshot from, must be in the same region 🤔 

> Something we could do is having a fixtures target which can be used to create a deployment with terraform. In any case, this would be an improvement on this PR as it would add a bunch of new concepts (fixtures, and the need for them when running some tests).

Sounds good, but yeah, it would require several big changes unrelated to this PR

Kushmaro:
@karencfv ,  the `source_cluster_id` field, is that the deployment id?
If so, it might make change to change terminology from "cluster" to "deployment"

Kushmaro:
> That's a very good question @ggsood. Yes, it will always restore the snapshot if a configuration change is made to the resource.

Hmm well, this doesn't make much sense (from a user's perspective), I get why it would do that.
Isn't there a way to "ignore" certain elements of a resource (IIRC there should be, but I'm really not sure)

if so, we should mark the snapshot field as such.
If not so.. I'm wondering if there's any other way to tackle this.

karencfv:
@Kushmaro thanks for the comment :) . Just so we're all on the same page I updated the description and the final fields look like this:

```hcl
resource "ec_deployment" "example_snapshot_source" {

  elasticsearch {
    snapshot_source {
      source_elasticsearch_cluster_id = "4f9c0cda4fe64b319a641ba6ee2689f7"
      snapshot_name = "__latest_success__"
    }
  }
}
```

PR #180, opened 2020-11-11
ec_deployment: Remove per topology config block

karencfv:
## Description
As the per topology user settings will be slowly phased out,
We will only be supporting global resource kind user settings
with the `config` block.

This patch removes all per topology `config` blocks.

## Motivation and Context
Supporting only what is necessary

## How Has This Been Tested?
acceptance tests

## Types of Changes

## Readiness Checklist


marclop (approved):
Yay for less settings! There's a merge conflict from my merged PR, but once that's resolved merge away

Issue #189, opened 2020-11-18
ec_deployment: Make node type fields editable


karencfv:
## Description
We shouldn’t block the users from setting node_types. We should set the defaults,
but allow customers the control they need in doing that.

PR #190, opened 2020-11-19
ec_deployment: Allow for ES node types to be editable

karencfv:
## Description
This patch gives users should the ability to set Elasticsearch node types.
All of the `node_type_*` arguments are now optional and computed.

## Related Issues
Closes: https://github.com/elastic/terraform-provider-ec/issues/189

## How Has This Been Tested?
Manually and through acceptance tests

## Types of Changes

## Readiness Checklist


docs/resources/ec_deployment.md
```diff
@@ -151,6 +151,10 @@ The optional `elasticsearch.topology` block supports the following arguments:
 * `size` - (Optional) Amount of memory (RAM) per topology element in the `"<size in GB>g"` notation. When omitted, it defaults to the deployment template value.
 * `size_resource` - (Optional) Type of resource to which the size is assigned. Defaults to `"memory"`.
 * `zone_count` - (Optional) Number of zones the instance type of the Elasticsearch cluster will span. This is used to set or unset HA on an Elasticsearch node type. When omitted, it defaults to the deployment template value.
+* `node_type_data` - (Optional) Node type (data) for the Elasticsearch cluster.
```
nrichers, line 154:
```suggestion
* `node_type_data` - (Optional) The data node type for the Elasticsearch cluster.
```

docs/resources/ec_deployment.md
```diff
@@ -151,6 +151,10 @@ The optional `elasticsearch.topology` block supports the following arguments:
...
+* `node_type_master` - (Optional) Node type (master) for the Elasticsearch cluster.
```
nrichers, line 155:
```suggestion
* `node_type_master` - (Optional) Node type for the Elasticsearch cluster (data node).
```

I'm not entirely sure these descriptions provide anything that you can't glean from the argument name, but I would perhaps put the text in brackets at the end to improve readability. Also, the [ES reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html#node-roles) refers to node _roles_ rather than types, but it seems our use of node types in descriptions is pretty pervasive to perhaps best to stick with what you have.

docs/resources/ec_deployment.md
```diff
@@ -151,6 +151,10 @@ The optional `elasticsearch.topology` block supports the following arguments:
...
+* `node_type_ingest` - (Optional) Node type (ingest) for the Elasticsearch cluster.
```
nrichers, line 156:
```suggestion
* `node_type_ingest` - (Optional) Node type for the Elasticsearch cluster (ingest node).
```

Similar comment to above.

docs/resources/ec_deployment.md
```diff
@@ -151,6 +151,10 @@ The optional `elasticsearch.topology` block supports the following arguments:
...
+* `node_type_ml` - (Optional) Node type (machine learning) for the Elasticsearch cluster.
```
nrichers, line 157:
```suggestion
* `node_type_ml` - (Optional) Node type for the Elasticsearch cluster (machine learning node).
```

Similar comment to above.

nrichers (approved):
LGTM, but I left some minor comments about improving readability.

docs/resources/ec_deployment.md
karencfv: node roles are a new field which we haven't implemented yet :)

ec/acc/testdata/deployment_basic_defaults_3.tf
```diff
@@ -12,6 +12,7 @@ resource "ec_deployment" "defaults" {
   elasticsearch {
     topology {
       size = "1g"
+      node_type_ingest = "false"
```
Kushmaro, line 15: @karencfv any reason on why we opt using strings here instead of boolean?

karencfv: Because booleans in Go default to false instead of nil. 
If we were to keep booleans, when a user wanted to use the default node types they would leave it blank and it would set to false. With strings we can check for an empty string and fill in with defaults if the string is empty.

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -130,6 +131,38 @@ func expandEsTopology(raw interface{}, topologies []*models.ElasticsearchCluster
 			}
 		}
 
+		if ntData, ok := topology["node_type_data"]; ok && ntData.(string) != "" {
+			nt, err := strconv.ParseBool(ntData.(string))
+			if err != nil {
+				return nil, err
```
marclop, line 137: Can we wrap this error with a meaningful string? I think it could end up being very confusing for users to receive the error that `strong.ParseBool` returns as is.

```suggestion
				return nil, fmt.Errorf("failed parsing node_type_data value: %w", err)
```

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -130,6 +131,38 @@ func expandEsTopology(raw interface{}, topologies []*models.ElasticsearchCluster
...
+			}
+			elem.NodeType.Data = ec.Bool(nt)
+		}
+
+		if ntMaster, ok := topology["node_type_master"]; ok && ntMaster.(string) != "" {
+			nt, err := strconv.ParseBool(ntMaster.(string))
+			if err != nil {
+				return nil, err
```
marclop, line 145:
```suggestion
				return nil, fmt.Errorf("failed parsing node_type_master value: %w", err)
```

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -130,6 +131,38 @@ func expandEsTopology(raw interface{}, topologies []*models.ElasticsearchCluster
...
+			}
+			elem.NodeType.Master = ec.Bool(nt)
+		}
+
+		if ntIngest, ok := topology["node_type_ingest"]; ok && ntIngest.(string) != "" {
+			nt, err := strconv.ParseBool(ntIngest.(string))
+			if err != nil {
+				return nil, err
```
marclop, line 153:
```suggestion
				return nil, fmt.Errorf("failed parsing node_type_ingest value: %w", err)
```

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -130,6 +131,38 @@ func expandEsTopology(raw interface{}, topologies []*models.ElasticsearchCluster
...
+			}
+			elem.NodeType.Ingest = ec.Bool(nt)
+		}
+
+		if ntMl, ok := topology["node_type_ml"]; ok && ntMl.(string) != "" {
+			nt, err := strconv.ParseBool(ntMl.(string))
+			if err != nil {
+				return nil, err
```
marclop, line 161:
```suggestion
				return nil, fmt.Errorf("failed parsing node_type_ml value: %w", err)
```

marclop (commented):
Approach looks good, left some improvements to provide more context on the bool parsing errors.

marclop (approved):

PR #194, opened 2020-12-02
all: Treat 404 and some 403 as resource destroyed

### LINKED ISSUE #193: all resources: Handle API 404 Not Found as destroyed resources
## Overview

Currently, we're not handling well when a resource managed by the provider is deleted, we're just returning the API error that's given to us from the SDK and not handling the case where the error is a 404 and sometimes 403 as well.

## Possible Implementation

Handle 404 and some 403 errors when the 403 is returned for security concerns.

## Context

Depends on https://github.com/elastic/cloud-sdk-go/issues/246

marclop:
## Description

Allows the provider to treat some 404 and 403 API responses to set the
resource that is managed by it as "gone" where it makes sense.

Adds proper handling of the "403" status code when the traffic filter
rule has already been deleted on on the update_traffic_rules file.

It also adds some unit tests to assert this behavior.

## Related Issues

Closes #193
Depends on elastic/cloud-sdk-go#258

## How Has This Been Tested?

Manually in multiple ways and with unit tests.

## Types of Changes


ec/ecresource/deploymentresource/read.go
```diff
@@ -67,3 +80,15 @@ func readResource(_ context.Context, d *schema.ResourceData, meta interface{}) d
...
+
+func deploymentNotFound(err error) bool {
+	// We're using the As() call since we do not care about the error value
+	// but do care about the error type since it's an implicit 404.
+	var notDeploymentNotFound *deployments.GetDeploymentNotFound
+	if errors.As(err, &notDeploymentNotFound) {
+		return true
+	}
+
+	// We also check for the case where a 403 is thrown for ESS.
```
karencfv, line 92: I realise ESS returns a 403 when a deployment doesn't exist, but I'm a bit concerned about the implications of doing this in the case that the user is getting a 403 for an actual authorisation issue 🤔

ec/ecresource/trafficfilterassocresource/read.go
```diff
@@ -45,3 +52,15 @@ func read(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diag
...
+func ruleAssocNotFound(err error) bool {
+	// We're using the As() call since we do not care about the error value
+	// but do care about the error type since it's an implicit 404.
+	var ruleNotFound *deployments_traffic_filter.GetTrafficFilterRulesetNotFound
+	if errors.As(err, &ruleNotFound) {
+		return true
+	}
+
+	// We also check for the case where a 403 is thrown for ESS.
+	return apierror.IsRuntimeStatusCode(err, 403)
```
karencfv, line 65:
```suggestion
	return apierror.IsRuntimeStatusCode(err, http.StatusForbidden)
```

ec/ecresource/trafficfilterassocresource/read.go
```diff
@@ -45,3 +52,15 @@ func read(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diag
...
+
+func ruleAssocNotFound(err error) bool {
+	// We're using the As() call since we do not care about the error value
+	// but do care about the error type since it's an implicit 404.
+	var ruleNotFound *deployments_traffic_filter.GetTrafficFilterRulesetNotFound
+	if errors.As(err, &ruleNotFound) {
+		return true
+	}
+
+	// We also check for the case where a 403 is thrown for ESS.
```
karencfv, line 64: same concern about checking for 403

ec/ecresource/trafficfilterresource/read.go
```diff
@@ -44,3 +51,15 @@ func read(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diag
...
+func ruleNotFound(err error) bool {
+	// We're using the As() call since we do not care about the error value
+	// but do care about the error type since it's an implicit 404.
+	var notDeploymentNotFound *deployments_traffic_filter.GetTrafficFilterRulesetNotFound
+	if errors.As(err, &notDeploymentNotFound) {
+		return true
+	}
+
+	// We also check for the case where a 403 is thrown for ESS.
+	return apierror.IsRuntimeStatusCode(err, 403)
```
karencfv, line 64:
```suggestion
	return apierror.IsRuntimeStatusCode(err, http.StatusForbidden)
```

ec/ecresource/trafficfilterresource/read.go
```diff
@@ -44,3 +51,15 @@ func read(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diag
...
+
+func ruleNotFound(err error) bool {
+	// We're using the As() call since we do not care about the error value
+	// but do care about the error type since it's an implicit 404.
+	var notDeploymentNotFound *deployments_traffic_filter.GetTrafficFilterRulesetNotFound
+	if errors.As(err, &notDeploymentNotFound) {
+		return true
+	}
+
+	// We also check for the case where a 403 is thrown for ESS.
```
karencfv, line 63: same

karencfv (commented):
The code looks good! I've only left a concern I have on the comments.

marclop:
I have gone over the possible cases where a 403 might be thrown and what the implications will be, from the testing I've performed this are the possible scenarios as I see them. I haven't been able to get the API to throw me a 403 when acting as a public API user.

The 403 status code indicates that an authenticated user is not authorized to access a given resource. A full definition:

> The HTTP 403 Forbidden client error status response code indicates that the server understood the request but refuses to authorize it.
>
> This status is similar to 401, but in this case, re-authenticating will make no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

I'm really unsure when the API would return a 403 if it would at all.

Bear in mind that all testing has been performed against the currently deployed ESS version (MS-48).

## Cases

### Deployment is shut down, not deleted

When the deployment is "Terminated" in the UI or shut down through the API (same thing), the deployment can still be accessed through: `GET /deployments/<id>`, but the resources appear stopped. This will never trigger a 403 or 404, we simply process this in the provider and `d.SetId("")` and return early without an error.

### Deployment is shut down and deleted

#### ECE

The API returns a 404, `d.SetId("")` and return early without an error.

#### ESS

I have never had the API return a 403 on a deployment that's created by the same user, due to the user's inability to delete a given deployment, thus we will always hit the first case described (Deployment is shut down, not deleted) in ESS.

<details>
<summary>Verbose log</summary>

```
$ ecctl --config ess deployment shutdown ef73c1f418162ddaa622d62b61024fbc -t
This action will delete the specified deployment ID and its associated sub-resources, do you want to continue? [y/n]: y
{
  "id": "ef73c1f418162ddaa622d62b61024fbc",
  "name": "my_example_deployment",
  "orphaned": {
    "apm": [],
    "appsearch": [],
    "elasticsearch": [
      {
        "dependents": [
          {
            "id": "dfda071940294497bb2ebf5233d33c33",
            "kind": "kibana"
          }
        ],
        "id": "8c15244ad39d46c6942f6a99793c2e8c"
      }
    ],
    "enterprise_search": [],
    "kibana": []
  }
}
Deployment [ef73c1f418162ddaa622d62b61024fbc] - [Elasticsearch][8c15244ad39d46c6942f6a99793c2e8c]: running step "perform-snapshot" (Plan duration 5.273251s)...
Deployment [ef73c1f418162ddaa622d62b61024fbc] - [Kibana][dfda071940294497bb2ebf5233d33c33]: running step "wait-until-stopped-ignore-allocator-status" (Plan duration 5.294255s)...
Deployment [ef73c1f418162ddaa622d62b61024fbc] - [Elasticsearch][8c15244ad39d46c6942f6a99793c2e8c]: running step "wait-until-stopped-ignore-allocator-status" (Plan duration 10.276262s)...
Deployment [ef73c1f418162ddaa622d62b61024fbc] - [Kibana][dfda071940294497bb2ebf5233d33c33]: running step "delete-instances" (Plan duration 10.297269s)...
Deployment [ef73c1f418162ddaa622d62b61024fbc] - [Elasticsearch][8c15244ad39d46c6942f6a99793c2e8c]: finished running all the plan steps (Total plan duration: 35.275891s)
Deployment [ef73c1f418162ddaa622d62b61024fbc] - [Kibana][dfda071940294497bb2ebf5233d33c33]: finished running all the plan steps (Total plan duration: 35.296894s)

$ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

ec_deployment.example_minimal: Refreshing state... [id=ef73c1f418162ddaa622d62b61024fbc]

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

$ ecctl --config ess deployment delete ef73c1f418162ddaa622d62b61024fbc --verbose
Using config file: /Users/marc/.ecctl/ess.yml
==================== Start of Request #0 ====================
DELETE /api/v1/deployments/ef73c1f418162ddaa622d62b61024fbc HTTP/1.1
Host: api.elastic-cloud.com
User-Agent: ecctl/v1.0.1
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip


====================  End of Request #0  ====================
==================== Start of Response #0 ====================
HTTP/1.1 404 Not Found
Content-Length: 67
Connection: keep-alive
Content-Type: application/json
Date: Thu, 03 Dec 2020 08:16:30 GMT
Server: fac/5c736b9
X-Request-Id: 946463ac66d4dc1992df083da540e57f

{"ok":false,"message":"The requested resource could not be found."}
====================  End of Response #0  ====================
[DELETE /deployments/{deployment_id}][404] deleteDeploymentNotFound  &{Errors:[]}

$ terraform plan

[Same result]

$ cat request.log
==================== Start of Request #1 ====================
GET /api/v1/deployments/ef73c1f418162ddaa622d62b61024fbc?convert_legacy_plans=false&enrich_with_template=true&show_metadata=true&show_plan_defaults=true&show_plan_history=false&show_plan_logs=false&show_plans=true&show_security=false&show_settings=true&show_system_alerts=5 HTTP/1.1
Host: api.elastic-cloud.com
User-Agent: elastic-terraform-provider/0.1.0-dev (cloud-sdk-go/2.8.0-ms48)
Accept: application/json
Authorization: [REDACTED]
Accept-Encoding: gzip


====================  End of Request #1  ====================
==================== Start of Response #1 ====================
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Connection: keep-alive
Content-Type: application/json
Date: Thu, 03 Dec 2020 08:17:39 GMT
Server: fac/5c736b9
X-Cloud-Resource-Created: 2020-12-03T08:09:55.537Z
X-Cloud-Resource-Last-Modified: 2020-12-03T08:14:11.050Z
X-Cloud-Resource-Version: 7
X-Ratelimit-Interval: 1 second
X-Ratelimit-Limit: 50
X-Ratelimit-Remaining: 49
X-Ratelimit-Reset: Thu, 03 Dec 2020 08:17:39 GMT
X-Request-Id: 2cb0957d83221a21dd2821df6da8cfea

[...]
```

</details>


## Conclusion

Seeing what we're seeing, I'm leaning towards removing the 403 case handling in the provider since users are unable to delete the deployments using the API, and in other types of installations ESSP or ECE they will get a 404 when the deployment truly can't be found.

marclop:
I've opened #195 to fix the `ec_deployments` datasource

karencfv:
Thanks for the thorough reply @marclop :) I am in agreement with you that it's best to not check for 403s as there is a possibility that the users will be getting a 403 for an actual authorisation issue.

That said, you will get a 403 if the deployment has been completely deleted. I _think_ @Kushmaro wanted to cover this case for disaster discovery. I would just like to double check we are all on the same page.

Regarding the traffic filter resource would there not be any implications of checking for a 403, given the possibility that the user might be hitting an actual authorisation error?

marclop:
@karencfv @Kushmaro I've added the 403 status code to be treated again as deployment is gone.

Additionally, I've fixed a long failing integration test where the traffic filter rule is gone, a deployment potentially needs to be updated and the traffic rule was `GET` again, a 403 was clearly returned (see previously failing integration tests) and now that the 403 is handled, the test is passing (finally!).

karencfv (approved):
Thanks for your patience with all the changes!

Issue #44, opened 2020-08-14
Publishing EC Terraform Provider to https://registry.terraform.io/


ggsood:
Plan to publish EC Terraform Provider to terraform registry.

As with Terraform 0.13 release providers can be downloaded directly from terraform registry.
Publishing the provider there will get more community visibility.

## Action items

- [x] Add `goreleaser` config and GitHub action to the project (#196).
- [x] Create a GitHub account to to release the provider on the Terraform Registry.
- [x] Create a GPG key to sign the release.
- [x] Add required secrets to GitHub secrets.
- [x] Push `v0.1.0-beta`
- [x] Publish the provider to the Terraform Registry.

marclop:
Thanks for opening this issue, we already had it on our radar and will work with Hashicorp to publish our provider

alejandrod:
This is a must :-) I hope it gets sorted out soon

akovtun-extenda:
Hello, @marclop. Any updates on this one?

marclop:
@akovtun-extenda we're busy working on having the first shippable version of the provider and publish it before the year closes, we're still uncertain on the exact release date.

arbourd:
@alejandrod has forked and published this package directly to the registry (very easy to do). This means you may have customers (I see 1000+ downloads) that are using an outdated fork instead of the official provider.

Would love to see the official source on the registry. 😬

marclop:
The terraform provider has been released and is available in the Terraform Registry as `0.1.0-beta`: https://registry.terraform.io/providers/elastic/ec/latest

tcarrondo:
@marclop  I'm getting:
`Error: Failed to query available provider packages

Could not retrieve the list of available versions for provider elastic/ec: no
available releases match the given constraints 0.1.0-beta`

Something as changed?

PR #196, opened 2020-12-04
release: Add goreleaser config and github action

marclop:
## Description

Adds the goreleaser config and release github action for pushes of `v*`
tags. Adds three new targets: `tag`, `snapshot` and `release`.
The `release` target has been added as a fallback to perform a manual
release in case the GitHub action fails.

Last, it sets the version to `v0.1.0-beta` as agreed with the team.

## Related Issues

Part of #44


karencfv (approved):
🎉

PR #208, opened 2020-12-15
ec_traffic_filter: Fix rule ordering bug

### LINKED ISSUE #203: ec_deployment_traffic_filter rules order
## Current Behavior
After creation ec_deployment_traffic_filter rules, every `terraform plan/apply` detects change due to rules order. 

## Possible Solution
Handling rules ordering.

## Steps to Reproduce
1. Crate ec_deployment_traffic_filter with min. two rules
2. Try to terraform apply/plan
3. Change is detected

## Your Environment
* Version used: 0.1.0-beta
* Operating System and version: macOS

marclop:
## Description

Fixes a bug where multiple having a traffic filter with a few rules will
cause an infinite diff loop by changing the field from a List to a Set.
This consistently hashes the list using the same algorithm, which does
not happen when the field is of `schema.TypeList`.

Also adds a custom set hashing function for the traffic rules set, using
the source and optional description if found as hashing keys.

Last, changes the assertions to use `TestCheckTypeSetElemNestedAttrs` in
the acceptance tests, asserting the rule values without the position.

## Related Issues

Closes #203

## How Has This Been Tested?

Unit and acceptance tested, also tried with an example locally.

## Types of Changes

### DIFF
--- a/ec/acc/deployment_traffic_filter_test.go
+++ b/ec/acc/deployment_traffic_filter_test.go
@@ -56,8 +56,12 @@ func TestAccDeploymentTrafficFilter_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resName, "include_by_default", "false"),
 					resource.TestCheckResourceAttr(resName, "type", "ip"),
 					resource.TestCheckResourceAttr(resName, "rule.#", "2"),
-					resource.TestCheckResourceAttr(resName, "rule.0.source", "1.1.1.0/24"),
-					resource.TestCheckResourceAttr(resName, "rule.1.source", "0.0.0.0/0"),
+					resource.TestCheckTypeSetElemNestedAttrs(resName, "rule.*", map[string]string{
+						"source": "0.0.0.0/0",
+					}),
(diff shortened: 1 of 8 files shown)


marclop:
The actual test that's touched here passes, there's a lot of noise currently in the acceptance tests with failures bubbling up from the platform due to ES instances failing to start in time:

From the Jenkins output:
```console
11:31:03  --- PASS: TestAccDeploymentTrafficFilter_basic (12.63s)
```

From my laptop:

```console
make testacc TEST_NAME=TestAccDeploymentTrafficFilter_basic
=== RUN   TestAccDeploymentTrafficFilter_basic
=== PAUSE TestAccDeploymentTrafficFilter_basic
=== CONT  TestAccDeploymentTrafficFilter_basic
--- PASS: TestAccDeploymentTrafficFilter_basic (17.29s)
PASS
ok  	github.com/elastic/terraform-provider-ec/ec/acc	17.707s
```

karencfv (approved):
That looks like a tricky bug to catch! Thanks for fixing it :)

PR #223, opened 2021-01-11
acc: fix ec_deployment post-destroy check

marclop:
## Description

Fixes the post-destroy check for the `ec_deployment` resource to only
error out when any of the deployments resources is still running and not
when the deployment is found on the backing API since it'll always be
there in ESS.

## Motivation and Context

Less acceptance test failures

## Types of Changes

### DIFF
--- a/ec/acc/deployment_destroy_test.go
+++ b/ec/acc/deployment_destroy_test.go
@@ -78,7 +78,7 @@ func testAccDeploymentDestroy(s *terraform.State) error {
 						continue
 					}
 					merr = merr.Append(
-						fmt.Errorf("deployment [%s] resource enterpriseSearch (%s) still exists",
+						fmt.Errorf("deployment [%s] resource enterprise_search (%s) still exists",
 							rs.Primary.ID, *res.ID,
 						),
 					)
--- a/ec/acc/deployment_destroy_test.go
+++ b/ec/acc/deployment_destroy_test.go
@@ -88,7 +88,7 @@ func testAccDeploymentDestroy(s *terraform.State) error {
 						continue
 					}
 					merr = merr.Append(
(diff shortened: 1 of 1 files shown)


ec/acc/deployment_destroy_test.go
```diff
@@ -60,31 +58,39 @@ func testAccDeploymentDestroy(s *terraform.State) error {
...
+						),
 					)
 				}
 				for _, res := range res.Resources.Kibana {
 					if *res.Info.Status == stoppedStatus {
 						continue
 					}
 					merr = merr.Append(
-						fmt.Errorf("resource apm (%s) still exists", *res.ID),
+						fmt.Errorf("deployment [%s] resource apm (%s) still exists",
```
karencfv, line 91:
```suggestion
						fmt.Errorf("deployment [%s] resource kibana (%s) still exists",
```

ec/acc/deployment_destroy_test.go
```diff
@@ -60,31 +58,39 @@ func testAccDeploymentDestroy(s *terraform.State) error {
...
+						),
 					)
 				}
 				for _, res := range res.Resources.EnterpriseSearch {
 					if *res.Info.Status == stoppedStatus {
 						continue
 					}
 					merr = merr.Append(
-						fmt.Errorf("resource enterpriseSearch (%s) still exists", *res.ID),
+						fmt.Errorf("deployment [%s] resource enterpriseSearch (%s) still exists",
```
karencfv, line 81:
```suggestion
						fmt.Errorf("deployment [%s] resource enterprise_search (%s) still exists",
```

karencfv: sorry! missed this one :P

karencfv (approved):

Issue #239, opened 2021-02-13
ec_deployments filter by name-prefix doesn't work with hyphens


oscarmorasu:
## Readiness Checklist

- [ x ] I am running the latest version
- [ ] I checked the documentation and found no answer
- [ ] I checked to make sure that this issue has not already been filed
- [ ] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

Given I have multiple Deployments with a hyphenated prefix `myorg-`:

```
myorg-cluster-dev
myorg-cluster-prod
myorg-monitoring
```

I should be able to find the Deployment called `myorg-monitoring` with this `name_prefix` filter:

```
data "ec_deployments" "example" {
  name_prefix = "myorg-monitoring"
}
```

## Current Behavior

The `ec_deployments` data source doesn't find any Deployments if the `name-prefix` filter contains hyphens

## Terraform definition

```
data "ec_deployments" "example" {
  name_prefix = "myorg-monitoring"
}
```

## Steps to Reproduce

1.
2.
3.
4.

## Context

I have several deployments prefixed with my organization name followed by a hyphen:

```
myorg-cluster-dev
myorg-cluster-prod
myorg-monitoring
```

I want to configure the dev and prod clusters to send the logs to the `myorg-monitoring` Deployment. I would like to leverage the `ec_deployments` to find the monitoring Deployment when provisioning the -dev and -prod Deployments.

## Your Environment

* Version used:
* Environment name and version (e.g. Go 1.9):
* Server type and version:
* Operating System and version:
* Link to your project:

karencfv:
Hi @oscarmorasu 👋 Thank you for finding this bug and taking the time to report it. I have opened up a PR with the fix.

PR #241, opened 2021-02-17
ec_deployments: Bugfix for queries with hyphens

karencfv:
## Description
Fixes a bug where the name prefix query was using a text field rather
than a keyword field in order to ensure the query is not analyzed.

Additionally the acceptance tests have been fixed since recently it
appears the deployment read API returns all possible topology elements
even if the size is set to 0. This means that the `ec_deployment` datasource
will return more topology elements than what is set in the `ec_deployment`
resource. I will open up a follow up PR with the fix for this, but I wanted to 
leave the tests passing for now.

## Related Issues
Closes: https://github.com/elastic/terraform-provider-ec/issues/239

## How Has This Been Tested?
Manually and through acceptance tests

## Types of Changes

## Readiness Checklist


marclop (approved):
Code LGTM

karencfv:
Will merge as the tests I was fixing passed

Screen Shot 2021-02-17 at 7 24 32 PM

PR #252, opened 2021-03-07
modified examples, removing patch version from TF requirements

Kushmaro:
## Description

- Added a configuration for ES nodes example to the basic deployment example.
- Removed .29 patch version from TF requirement, as it confused users to our minimum supported version (which should really be 0.12.X
- apm / enterprisesearch blocks gave me an error upon testing (i think mainly due to plan changes ) opted to remove them altogether

## Motivation and Context
Wanted to add an "es_settings.yaml" usage example since got asked by field. :) 
Got rid of .29 along the way

## How Has This Been Tested?
Ran locally

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/examples/deployment/deployment.tf
+++ b/examples/deployment/deployment.tf
@@ -9,9 +9,7 @@ terraform {
   }
 }
 
-provider "ec" {
-  apikey="TmxIdURIZ0JxVld3ME5DWTdqVUM6WnhTWEZoWXpRRmVMMXh6WHQ0U2RIQQ=="
-}
+provider "ec" {}
 
 # Create an Elastic Cloud deployment
 resource "ec_deployment" "example_minimal" {
--- a/examples/deployment/deployment.tf
+++ b/examples/deployment/deployment.tf
@@ -1,15 +1,17 @@
 terraform {
-  required_version = ">= 0.12.29"
+  required_version = ">= 0.12"
 
   required_providers {
     ec = {
       source  = "elastic/ec"
-      version = "0.1.0-beta"
+      version = ">=0.1.0-beta"
     }
   }
 }
 
-provider "ec" {}
+provider "ec" {
+  apikey="TmxIdURIZ0JxVld3ME5DWTdqVUM6WnhTWEZoWXpRRmVMMXh6WHQ0U2RIQQ=="
+}
 
 # Create an Elastic Cloud deployment
 resource "ec_deployment" "example_minimal" {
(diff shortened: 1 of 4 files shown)


examples/deployment/deployment.tf
```diff
@@ -1,10 +1,10 @@
 terraform {
-  required_version = ">= 0.12.29"
+  required_version = ">= 0.12"
```
karencfv, line 2: This was set to `">= 0.12.29"` because there are some changes which were fixed in that version as far as I remember. We should probably keep it like it was :) If this is confusing people then maybe you can bump it to `>=0.13` ?

examples/deployment/deployment.tf
```diff
@@ -21,19 +21,11 @@ resource "ec_deployment" "example_minimal" {
...
+    config {
+	user_settings_yaml = file("./es_settings.yaml")
     }
   }
 
-  enterprise_search {
-    topology {
-      zone_count = 1
-    }
-  }
```
karencfv, line 38: The idea with these two was to show the different blocks there are for a deployment resource and how to use the topology element block. I think it's a good idea to keep them. I'm not sure what happened before but I tried the example myself and all the resources were created successfully :)

examples/deployment/deployment.tf
```diff
@@ -21,19 +21,11 @@ resource "ec_deployment" "example_minimal" {
...
-}
```
karencfv, line 39: The linter is failing on this file :( It might be because you added a newline at the end of the file?

examples/deployment_ccs/deployment.tf
```diff
@@ -63,4 +63,4 @@ resource "ec_deployment" "ccs" {
   }
 
   kibana {}
-}
```
karencfv, line 66: same here with the new line

examples/deployment/deployment.tf
Kushmaro: Interesting, I wasn't aware that we had issues in <0.12.29; 
does the provider not work at all under that version?

I'd just want us to not create too strong of a "support" message here. It may also be that the first thing we do when people see things not working properly on <0.12.29 is to upgrade to the latest 0.12.X

Just a thought. Anyways, i'd be happy to know what sort of issues were seen in earlier versions so that I have more context.

examples/deployment/deployment.tf
Kushmaro: sigh that linter! i'll fix.

examples/deployment/deployment.tf
Kushmaro: Let me re-test this.

examples/deployment/deployment.tf
marclop: Yeah, we need `0.12.29` for terraform to not crash when `terraform` block is set.

PR #253, opened 2021-03-08
ec_deployment: add support for data tiers

### LINKED ISSUE #212: ec_deployment: Support node_roles and data tiers
## Overview

We need to add support for the upcoming `node_roles` field that will be introduced soon.

## Possible Implementation

There's a few things to consider:

* New `id` field will be the unique selector for topology elements, no longer `instance_configuration_id`.
* Migration to `node_roles` from `node_type_*` when version is >= `7.11.0`.
* Leave `node_type_*` as they are when the version is lower than `7.11.0`.

We should be mindful of how this change affects the current boolean `node_type_*` on the state and how to treat potential changes. Caveats should be documented.

marclop:
## Description

I've split the pull request in individual commits which should make it
easier to review.

Adds support for `node_roles` and automatic migration from `node_type_*`
to `node_roles`. I will note a few things to make it easier to break
down all of these changes.

f972aa8:
Adds `node_roles` support by updating the SDK to the latest available
version, this makes the deployment template not work out of the box and
instead our client has to choose whether to use the `node_type` object
or the `node_roles` array.
Stack versions >= `7.10.0` can and should use `node_roles` since that
allows users to leverage the `cold` tier and autoscaling in the future
as well.
With the introduction of the `"id"` field, the topology elements needn't
use `instance_configuration_id` as the unique field, instead, `id` is
that unique field value.
**The `instance_configuration_id` has been set as computed only** and
cannot be set by the user anymore. For now, this change is present on
the Elasticsearch topology only, in the future, as other applications
might also auto scale, the API will change accordingly.
Last, it validates that a set topology element has a size > 0, otherwise
an error will be thrown.

99dfe21:
Another somewhat breaking change is the removal of the version computed
property in the Elasticsearch object. This field is completely redundant
and should not be used, instead the global version declares the version
for all of the deployment's applications. Since we're removing a field
from the schema and previous versions of the provider have saved it in
the state, we're adding a `StateUpgradeFunc` and bumping the current
schema version to `1`. The schema upgrade function will only be run when
the saved state version is `0` and will only run once, then the state
schema version will be bumped to 1.

The rest of commits are updates to the tests, fixtures, documentation
and example files.

## Related Issues
Closes #212 

## How Has This Been Tested?

Thoroughly unit and acceptance tested, additionally I've tested the same scenarios as the acceptance tests manually.

## Types of Changes

- [x] Documentation


docs/resources/ec_deployment.md
```diff
@@ -166,18 +202,17 @@ To set up multi-node Elasticsearch clusters, you can set single or multiple topo
 
 The optional `elasticsearch.topology` block supports the following arguments:
 
-* `instance_configuration_id` - (Optional) Default instance configuration of the deployment template. To change it, use the [full list](https://www.elastic.co/guide/en/cloud/current/ec-regions-templates-instances.html) of regions and deployment templates available in ESS.
-
--> Before you get started with instance configurations, read the [ESS hardware and Instance Configurations](https://www.elastic.co/guide/en/cloud/current/ec-reference-hardware.html#ec-instance-configuration-names) documentation.
-
-* `size` - (Optional) Amount of memory (RAM) per topology element in the `"<size in GB>g"` notation. When omitted, it defaults to the deployment template value.
+* `id` - (Required) Unique topology identifier, this will generally refer to an Elasticsearch data tier, such as `hot_content`, `warm` or `cold`.
```
alaudazzi, line 205:
```suggestion
* `id` - (Required) Unique topology identifier. It generally refers to an Elasticsearch data tier, such as `hot_content`, `warm` or `cold`.
```

docs/resources/ec_deployment.md
```diff
@@ -166,18 +202,17 @@ To set up multi-node Elasticsearch clusters, you can set single or multiple topo
...
 * `size_resource` - (Optional) Type of resource to which the size is assigned. Defaults to `"memory"`.
 * `zone_count` - (Optional) Number of zones the instance type of the Elasticsearch cluster will span. This is used to set or unset HA on an Elasticsearch node type. When omitted, it defaults to the deployment template value.
 * `node_type_data` - (Optional) The node type for the Elasticsearch cluster (data node).
 * `node_type_master` - (Optional) The node type for the Elasticsearch cluster (master node).
 * `node_type_ingest` - (Optional) The node type for the Elasticsearch cluster (ingest node).
 * `node_type_ml` - (Optional) The node type for the Elasticsearch cluster (machine learning node).
 
+~> **Note when node_type_* fields set** After upgrading to a version that supports data tiers (7.10.0 or above), the `node_type_*` will not have any effect even if specified. The provider will automatically migrate the `node_type_*` fields to the appropriate `node_roles` as set by the deployment template. After having upgraded to `7.10.0` or above, the fields should be removed from the terraform configuration if explicitly configured.
```
alaudazzi, line 214:
```suggestion
~> **Note when node_type_* fields set** After upgrading to a version that supports data tiers (7.10.0 or above), the `node_type_*` has no effect even if specified. The provider automatically migrates the `node_type_*` fields to the appropriate `node_roles` as set by the deployment template. After having upgraded to `7.10.0` or above, the fields should be removed from the terraform configuration, if explicitly configured.
```

docs/resources/ec_deployment.md
```diff
@@ -308,10 +343,12 @@ In addition to all the arguments above, the following attributes are exported:
 * `elasticsearch.#.cloud_id` - Encoded Elasticsearch credentials to use in Beats or Logstash. For more information, see [Configure Beats and Logstash with Cloud ID](https://www.elastic.co/guide/en/cloud/current/ec-cloud-id.html).
 * `elasticsearch.#.http_endpoint` - Elasticsearch resource HTTP endpoint.
 * `elasticsearch.#.https_endpoint` - Elasticsearch resource HTTPs endpoint.
+* `elasticsearch.#.topology.#.instance_configuration_id` - instance configuration of the deployment topology element.
 * `elasticsearch.#.topology.#.node_type_data` - Node type (data) for the Elasticsearch topology element.
 * `elasticsearch.#.topology.#.node_type_master` - Node type (master) for the Elasticsearch topology element.
 * `elasticsearch.#.topology.#.node_type_ingest` - Node type (ingest) for the Elasticsearch topology element.
 * `elasticsearch.#.topology.#.node_type_ml` - Node type (machine learning) for the Elasticsearch topology element.
+* `elasticsearch.#.topology.#.node_roles` - List of roles for the the topology element. They are inferred from the deployment template.
```
alaudazzi, line 351:
```suggestion
* `elasticsearch.#.topology.#.node_roles` - List of roles for the topology element. They are inferred from the deployment template.
```

alaudazzi (approved):
Left a few editing suggestions. LGTM for the documentation part.

karencfv (commented):
This is looking great @marclop 🎉 ! Just wondering if this PR should include an update to the `ec_deployment` data source as well 🤔 shouldn't be too big of a change I think? Otherwise I guess that should be tackled as a follow up

ec/acc/deployment_ccs_test.go
```diff
@@ -63,10 +63,12 @@ func TestAccDeployment_ccs(t *testing.T) {
...
-					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_data", "true"),
-					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_ingest", "true"),
-					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_master", "true"),
-					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_ml", "false"),
+					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_data", ""),
+					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_ingest", ""),
+					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_master", ""),
+					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.node_type_ml", ""),
+					resource.TestCheckResourceAttr(ccsResName, "elasticsearch.0.topology.0.id", "hot_content"),
+					resource.TestCheckResourceAttrSet(ccsResName, "elasticsearch.0.topology.0.node_roles.#"),
```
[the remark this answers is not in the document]
marclop: I thought about it, but `node_roles` are likely to evolve over time, if we're too strict in our assertions we'll end up having a lot of maintenance overhead, so I've opted to at least for now check this lightly rather than assert all the node roles for all topology elements.

marclop:
@karencfv I'll add the `datasource` schema change as well.

karencfv (approved):
Thanks for adding this to the data source as well!

marclop:
Tests actually passed, the post destroy tracking is what returned an error, so we're good to merge.

PR #275, opened 2021-03-25
acceptance: Increase timeout on API calls

karencfv:
## Description
Increases the timeout value to 5 minutes on acceptance tests to
hopefully give enough time for the destroy actions to finish.

## Motivation and Context
Make the tests pass again 😩 

## How Has This Been Tested?
We'll see what happens when Jenkins runs

## Types of Changes

## Readiness Checklist

### DIFF
--- a/ec/acc/acc_prereq.go
+++ b/ec/acc/acc_prereq.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/auth"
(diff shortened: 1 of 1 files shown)


ec/acc/acc_prereq.go
```diff
@@ -118,6 +119,7 @@ func newAPI() (*api.API, error) {
 		Host:          host,
 		SkipTLSVerify: insecure,
 		Retries:       ec.DefaultHTTPRetries,
+		Timeout:       5 * time.Minute,
```
marclop, line 122: This won't have the desired effect. The timeout here controls the http client timeout before it the Go `http.Client` instance cancels an individual the request.

To make matters worse, there's no timeout parameter in the shutdown operation so we can't work around it on the client side: https://github.com/elastic/cloud-sdk-go/blob/master/api/apidocs-user.json#L1731-L1756

PR #300, opened 2021-04-21
ec_deployments: Add size parameter to data source 

### LINKED ISSUE #294: ec_deployments : Expose size field for search requests
## Overview

The current `modelToState` function iterates over the deployments returned from the API search endpoint (in `DeploymentsSearchResponse`). The `SearchRequest` is defaulting to `size:10` , and only 10 deployments are returned in the response.

## Possible Implementation

~Use the `match_count` and `return_count` fields in `DeploymentsSearchResponse`, alongside the `from` field in `SearchParams` to page over the possible results.~

After chatting to @marclop we realised that pagination would mutate the state for every page that was iterated over. A better/easier solution is to expose the `size` field (with a default of 10) and allow the user to set an upper limit.

marclop:
## Description

Adds a new `size` integer parameter to the `ec_deployments` data source
which allows users to specify a maximum number of deployments to be
returned. The API defaults to 10 and the provider defaults to `100`.

## Related Issues

Closes #294 

## Types of Changes

### DIFF
--- a/.changelog/300.txt
+++ b/.changelog/300.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+datasource/ec_deployment: Add a new size parameter to allow modifying the default size of `10` in the number of deployments returned by the search request.
+```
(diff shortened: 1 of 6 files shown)


karencfv (approved):

alaudazzi (approved):
LGTM

neiljbrookes (approved):
Thanks @marclop this looks great ✅

PR #313, opened 2021-05-14
Use `elasticcloudclients` token to checkout

marclop:


marclop:
manually aborted the acceptance build.

PR #322, opened 2021-06-07
ec_deployments: Use API sorting by deployment ID

### LINKED ISSUE #320: Consistent sort order in datasource
We use ec_deployments datasource to populate a list of remote_clusters in a different deployment.

Every time we do a plan/apply it looks like there is a change, but it's not, the list of deployments is just in a different order.

I think it can be fixed by specifying the sort parameter in the request
https://github.com/elastic/terraform-provider-ec/blob/2473642a2fc5c51f2e27b98f7699145734fbf0e6/ec/ecdatasource/deploymentsdatasource/expanders.go#L94-L95
https://github.com/elastic/cloud-sdk-go/blob/e6d054f02355d17805dd852166082c2b0e322544/pkg/models/search_request.go#L48
If we sort by e.g. deployment_it should make it stable and idempotent.

marclop:
## Description

Uses the `Sort` field of the Search body request to always sort the
resulting deployments by ID, since they were not, now it should ALWAYS
return the same result order as long as the deployment array doesn't
change.

## Related Issues

Closes #320

## Types of Changes

### DIFF
--- a/.changelog/322.txt
+++ b/.changelog/322.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+datasource/ec_deployments: Properly sorts the datasource results by ID.
+```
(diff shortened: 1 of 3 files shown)


alpar-t (approved):

karencfv (approved):
As the failing test is unrelated to the TF provider, happy to have this merged

PR #350, opened 2021-08-19
docs: Add `RELEASE.md` document

marclop:
## Description

Adds a new `RELEASE.md` document and updates the CONTRIBUTING doc to add
changelog entries for every change.

### DIFF
--- a/developer_docs/RELEASE.md
+++ b/developer_docs/RELEASE.md
@@ -1,0 +1,58 @@
+# Releasing a new version
+
+This guide aims to provide guidance on how to release new versions of `terraform-provider-ec`.
+
+- [Releasing a new version](#releasing-a-new-version)
+  - [Prerequisites](#prerequisites)
+    - [Make sure `VERSION` is up to date](#make-sure-version-is-up-to-date)
+    - [Ensure the `elastic/cloud-sdk-go` dependency is up to date](#ensure-the-elasticcloud-sdk-go-dependency-is-up-to-date)
+    - [Generating a changelog for the new version](#generating-a-changelog-for-the-new-version)
+      - [Patch version changelog](#patch-version-changelog)
+  - [Executing the release](#executing-the-release)
+  - [Post release tasks](#post-release-tasks)
+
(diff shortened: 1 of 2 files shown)


developer_docs/RELEASE.md
```diff
@@ -0,0 +1,58 @@
...
+
+Releasing a new version implies that there have been changes in the source code which are meant to be released for wider consumption. Before releasing a new version there's some prerequisites that have to be checked.
+
+### Make sure `VERSION` is up to date
+
+**Since the `VERSION` is now updated via github actions, just double check that it is updated, and if not, manually do so**.
+
+Since the source has changed, we need to update the current committed version to a higher version so that the release is published.
+
+The version is currently defined in the [Makefile](./Makefile) as an exported environment variable called `VERSION` in the [SEMVER](https://semver.org) format: `MAJOR.MINOR.BUG`
```
jeredding, line 24: `MAJOR.MINOR.BUG` -> `MAJOR.MINOR.PATCH` bc `If a patch version needs to be released` down below.

developer_docs/RELEASE.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+Say we want to perform a minor version release (i.e. no breaking changes and only new features and bug fixes are being included); in which case we'll update the _MINOR_ part of the version, this can be done with the `make minor` target, but it should have been updated automatically via GitHub actions.
+
+```Makefile
+SHELL := /bin/bash
+export VERSION ?= v1.1.0
+```
+
+If a patch version needs to be released, the release will be done from the minor branch. For example, if we want to release `v0.2.1`, we will check out the `0.2` branch and perform any changes in that branch. The VERSION variable in the Makefile should already be up to date, but in case it's not, it can be bumped with the `make patch` target.
```
jeredding, line 38: see above comment on semver definition.

jeredding (changes requested):
one thing about `patch` otherwise looks great

jeredding (approved):
LGTM

PR #370, opened 2021-09-06
ec_deployment: Add `config.docker_image` argument

marclop:
## Description

Adds a new `config.docker_image` argument to all resources which allows
Elastic Cloud internal users to override the docker image used.

**This is an internal only field, and attempting to set it to any value
when you're not part of the Elastic org, will result in an API error**.

## Motivation and Context

Better support for docker image overrides.

## How Has This Been Tested?

Unit and acceptance tested 

## TODO:

- [x] Add acceptance tests using a durable docker image. (b7d915b)
- [x] Update documentation to reflect this field, pending on Cloud Writers.


marclop:
Code is in a reviewable state, but still waiting for guidance on how to document this from @elastic/cloud-writers

obierlaire (approved):

PR #371, opened 2021-09-06
ec_deployment: Persist lowest resource `version`

marclop:
## Description

Reconciles the top level `ec_deployment.version` field to the lesser of
all of the resource versions. Cases where the version is updated and one
or more of the resources did not succeed to upgrade, were not detected
since the `version` field was not reconciled during the Read stage and
was set to the value the terraform configuration was set to without the
ability to detect drift.

## Motivation and Context

Failed upgrades could not be retried through the provider.

## How Has This Been Tested?

Unit and acceptance tested.

## Types of Changes

### DIFF
--- a/.changelog/371.txt
+++ b/.changelog/371.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/ec_deployment: Fix a bug affecting partial version upgrades, where some resources upgraded but others did not, leaving the `version` argument set to the configured value as if the upgrade was successful. The upgrade could not be retried and it was unable to detect drift.
+```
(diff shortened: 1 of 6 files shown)


.changelog/371.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/ec_deployment: Fix a bug affecting partial version upgrades, where some resources upgraded but others did not, leaving the `version` argument set to the configured value as if the upgrade was successful. The upgrade could not be retried and it was unable to detect drift.
```
marclop, line 2: I'm not so sure on the wording here, we might need a wordsmith to come in and fix this strange sentence. @alaudazzi if you have some minutes 🙂

obierlaire (approved):

.changelog/371.txt
kilfoyle, line 2:
```suggestion
resource/ec_deployment: Fixed a bug that affects partial version upgrades. During an upgrade only a subset of resources would upgrade successfully, but the `version` argument value updated as if all resources were upgraded. Attempts to retry the upgrade would fail since the version difference was not detected.
```
I think Arianna is offline at the moment so I'd suggest the above for the bug text.

kilfoyle (approved):
LGTM for the text updates. 👍

.changelog/371.txt
marclop: Thanks!

PR #372, opened 2021-09-06
Add auto-commit of NOTICE file

artemnikitin:
## Description

This PR adds a Github Action to commit changed NOTICE file

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/.github/workflows/changelog_generation.yml
+++ b/.github/workflows/changelog_generation.yml
@@ -26,9 +26,7 @@ jobs:
           if [[ ! -z  $(git status --porcelain) ]]; then
             git config --local user.email elasticcloudclients@elastic.co
             git config --local user.name elasticcloudclients
-            git add CHANGELOG.md
-            git commit -m "Update CHANGELOG.md"
-            git add NOTICE
-            git commit -m "Update NOTICE"
+            git add CHANGELOG.md NOTICE
+            git commit -m "Update CHANGELOG.md and NOTICE"
             git push
           fi
--- a/.github/workflows/changelog_generation.yml
+++ b/.github/workflows/changelog_generation.yml
@@ -1,4 +1,4 @@
-name: Generate CHANGELOG
+name: Generate CHANGELOG and NOTICE
 on:
   push:
     branches:
@@ -20,7 +20,7 @@ jobs:
           go-version: 1.17
         id: go
 
-      - run: make changelog
+      - run: make changelog vendor
       - name: Commit & Push changes
         run: |
           if [[ ! -z  $(git status --porcelain) ]]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -46,3 +46,8 @@ jobs:
     - name: Validate the examples
       run: make install validate-examples
       id: validate-examples
+
+    - name: Commit generated files
+      uses: stefanzweifel/git-auto-commit-action@v4
```
marclop, line 51: We're using a different action on other workflows: https://github.com/elastic/terraform-provider-ec/blob/master/.github/workflows/branch.yml#L44-L51. Any reason not to re-use the same one here?

artemnikitin: Thank you for the suggestion, Marc 👍 

> Any reason not to re-use the same one here?

Not at all, I just missed it. I will change my PR to use this one

artemnikitin: Fixed

marclop:
change looks good, could you try to change the NOTICE file (Remove some lines or so) in a commit to see if the GitHub action commits it back?

.github/workflows/go.yml
artemnikitin: > We're using a different action on other workflows: https://github.com/elastic/terraform-provider-ec/blob/master/.github/workflows/branch.yml#L44-L51. 

I switched to it, but having an issue similar to https://github.com/EndBug/add-and-commit/issues/199. Trying to fix it...

marclop:
It may be easier to just add the NOTICE file generation just like we generate and commit files in the repository branches (`master` and version branches `x.x`). Seems like that GitHub action is not very happy to commit on forked repositories.

.github/workflows/changelog_generation.yml
```diff
@@ -20,13 +20,15 @@ jobs:
...
+      - run: make changelog vendor
       - name: Commit & Push changes
         run: |
           if [[ ! -z  $(git status --porcelain) ]]; then
             git config --local user.email elasticcloudclients@elastic.co
             git config --local user.name elasticcloudclients
             git add CHANGELOG.md
             git commit -m "Update CHANGELOG.md"
+            git add NOTICE
+            git commit -m "Update NOTICE"
```
marclop, line 32: I'm not sure this approach will work very well when there aren't any changelog updates but we need to re-generate the NOTICE file. It's probably not a big deal, but this means that the NOTICE file will only be ever updated when the changelog has new entries.

```console
$ git commit -m "test"; echo $?
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
1
```

I would instead do add both files on the `git add` on line 29 and commit both files with an updated commit message.

artemnikitin: Fixed

marclop (approved):
Awesome, Thank you for all the work Artem. I know testing this wasn't the easiest!

PR #364, opened 2021-09-02
ec_deployment_elasticsearch_keystore: Add resource

marclop:
## Description

Adds a new `ec_deployment_elasticsearch_keystore` resource which allows
allows creating and updating Elasticsearch keystore settings.

Elasticsearch keystore settings can be created and updated through this
resource, each resource represents a single Elasticsearch Keystore
setting. After adding a key and its secret value to the keystore, you
can use the key name in place of the secret value when you configure
sensitive settings.

This resource offers weaker consistency guarantees and will not detect
and update keystore setting values that have been modified outside of
the scope of Terraform, usually referred to as _drift_. For instance,
examine the following scenario:
    1. A keystore setting is created using this resource.
    2. The keystore setting's value is modified to a different value
       using the Elasticsearch Service API.
    3. Running `terraform apply` will fail to detect the changes and
       will not update the keystore setting to the value defined in the
       terraform configuration.

To force the keystore setting to the value it is configured to hold,
you may want to taint the resource and force its recreation.

Thanks to @karencfv and @Crazybus for the initial work.

## Related Issues

Closes #100

## How Has This Been Tested?

Manually, unit and acceptance tested.

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)

PR #382, opened 2021-09-13
Add changelog entry for keystore resource

marclop:
## Description

Adds the missing changelog entry for the change introduced in #364


marclop:
cancelled acceptance tests.

PR #440, opened 2022-02-10
Not subject to support SLA

ppf2:
Documenting support SLA expectations for the provider.  Feel free to reword as needed :)

### DIFF
--- a/docs-elastic/index.asciidoc
+++ b/docs-elastic/index.asciidoc
@@ -6,6 +6,6 @@ The {n} allows you to provision Elastic Cloud deployments on any Elastic Cloud p
 
 The provider lets you manage Elastic Cloud deployments as code, and introduce DevOps-driven methodologies to manage and deploy the Elastic Stack and solutions.
 
-IMPORTANT: The provider is not subject to the support service level agreement (SLA.) You can use the Elastic Cloud APIs or user console directly to perform changes as a workaround.
+IMPORTANT: The provider is not subject to the support service level agreement (SLA.) You can use the [Elastic Cloud APIs](https://www.elastic.co/guide/en/cloud/current/ec-restful-api.html) or [Elasticsearch Service Console](https://www.elastic.co/guide/en/cloud/current/ec-configure-settings.html) directly to perform changes as a workaround.
 
 To get started, see our link:https://registry.terraform.io/providers/elastic/ec/latest/docs[Elastic Cloud Terraform provider documentation].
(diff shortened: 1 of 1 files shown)


docs-elastic/index.asciidoc
```diff
@@ -6,4 +6,6 @@ The {n} allows you to provision Elastic Cloud deployments on any Elastic Cloud p
 
 The provider lets you manage Elastic Cloud deployments as code, and introduce DevOps-driven methodologies to manage and deploy the Elastic Stack and solutions.
 
+NOTE: The provider is not subject to the support SLA. Users can use the Elastic Cloud APIs or user console directly to perform changes as a workaround.
```
nrichers, line 9:
```suggestion
IMPORTANT: The provider is not subject to the support service level agreement (SLA.) You can use the Elastic Cloud APIs or user console directly to perform changes as a workaround.
```

Can you add some links here, e.g. to the user console or API docs? I didn't suggest them as I don't actually know what we support for the Terraform provider. (If you'd rather I add the links, just comment here with what they are.)

nrichers (approved):
LGTM, just a small consistency edit that might help. If you have links to include, please do? (It might also help to add a note into the actual Terraform provider docs that this page points to, but AFAIK the only two developers to ever have touched those docs directly are no longer at Elastic, meaning some additional legwork for process discovery might be required.)

For my own reference, this PR will need to be backported into `0.2` and `0.1` as per https://github.com/elastic/docs/blob/master/conf.yaml#L954.

docs-elastic/index.asciidoc
```diff
@@ -6,4 +6,6 @@ The {n} allows you to provision Elastic Cloud deployments on any Elastic Cloud p
 
 The provider lets you manage Elastic Cloud deployments as code, and introduce DevOps-driven methodologies to manage and deploy the Elastic Stack and solutions.
 
+IMPORTANT: The provider is not subject to the support service level agreement (SLA.) You can use the [Elastic Cloud APIs](https://www.elastic.co/guide/en/cloud/current/ec-restful-api.html) or [Elasticsearch Service Console](https://www.elastic.co/guide/en/cloud/current/ec-configure-settings.html) directly to perform changes as a workaround.
```
ppf2, line 9: Added links, WDTY? @nrichers

tobio, line 9: > directly to perform changes as a workaround

For many people this is unlikely to be true. Terraform is often setup as the _only_ way changes can get into production. Intentionally introducing drift from the desired state (as defined in the Terraform resources) can also be a pretty big risk for managed resources, reconciling that drift can easily create outages. I don't think we should casually direct people down this path. 

What's the reason this client is being carved out of the SLA and not Cloud UI?

ppf2: Thanks for the feedback! This PR is the outcome from an internal discussion thread.  I have looped you in so we can continue to refine the arrangement.

ppf2:
Closing this PR. We will continue the discussion and will resubmit the PR with the appropriate statement when ready.

Issue #443, opened 2022-02-23
TestAccDeployment_snapshot_restore is flaky


marclop:
## Overview

The test is flaky and failing quite often, we should investigate why it's failing and fix the test.

```
=== CONT  TestAccDeployment_snapshot_restore
    deployment_snapshot_test.go:104: PUT https://8f015c78eccd4935a980b51a34b62d1a.us-east-1.aws.found.io:9243/_snapshot/found-snapshots/snap_1645544308?wait_for_completion=true
    deployment_snapshot_test.go:117: Snapshot created
--- PASS: TestAccDeployment_basic_tags (444.40s)
=== CONT  TestAccDeployment_pre_node_roles
=== CONT  TestAccDeployment_snapshot_restore
    deployment_snapshot_test.go:39: Step 2/3 error: Error running apply: exit status 1
        
        Error: failed tracking create progress: 2 errors occurred:
        	* found deployment plan errors: deployment [141191a01df9a6fc33e1fdf1847cda87] - [elasticsearch][3e7e996fa5644509ae10ed0670e63c44]: caught error: "Plan change failed: Could not restore snapshot: received error response from Elasticsearch restore API."
        	* set "request_id" to "ztbonl15ckiotp7vnw241e66c7ofhu1e3dpk5o5pzxgcd9af9llbcjq7mj0ttwlm" to recreate the deployment resources
        
        
        
          with ec_deployment.snapshot_target,
          on terraform_plugin_test.tf line 25, in resource "ec_deployment" "snapshot_target":
          25: resource "ec_deployment" "snapshot_target" {
        
```

PR #444, opened 2022-02-23
acceptance: Skip Deployment_snapshot_restore test

marclop:
## Description

Skips the test for now since it seems to be failing quite often and it
makes it difficult to rely on the acceptance tests for PR checks.

## Related Issues

Part of #443


mieciu (approved):
**LGTM**, although we've run into few other failures 😅

marclop:
Merging since the acceptance tests will be fixed in #425

Issue #471, opened 2022-04-19
Terraform provider does no handle unsized but enabled ML tiers


tobio:
## Readiness Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

Managing the following resources should create the topology elements once, and then not show any required changed on a subsequent `terraform plan`

```hcl
resource "ec_deployment" "staging" {
  region                 = "eu-west-1"
  version                = data.ec_stack.latest.version
  deployment_template_id = "aws-storage-optimized"

  elasticsearch {
    autoscale = "true"

    topology {
      id         = "hot_content"
      size       = "1g"
      zone_count = 3

      autoscaling {
        max_size = "8g"
      }
    }

    topology {
      id         = "ml"
      zone_count = 1

      autoscaling {
        min_size = "0gb"
        max_size = "60gb"
      }
    }
  }

  kibana {
    topology {
      size       = "1g"
      zone_count = 2
    }
  }
}
```

## Current Behavior
The provider does not treat the unsized ML tier as existing, and attempts to create it on each subsequent `terraform plan|apply`

## Terraform definition

```hcl
resource "ec_deployment" "staging" {
  region                 = "eu-west-1"
  version                = data.ec_stack.latest.version
  deployment_template_id = "aws-storage-optimized"

  elasticsearch {
    autoscale = "true"

    topology {
      id         = "hot_content"
      size       = "1g"
      zone_count = 3

      autoscaling {
        max_size = "8g"
      }
    }

    topology {
      id         = "ml"
      zone_count = 1

      autoscaling {
        min_size = "0gb"
        max_size = "60gb"
      }
    }
  }

  kibana {
    topology {
      size       = "1g"
      zone_count = 2
    }
  }
}
```

## Steps to Reproduce

1. Use the above resource
2. `terraform apply`
3. `terraform apply`

PR #472, opened 2022-04-20
Allow zero sized topology elements when autoscaling is enabled

tobio:
## Related Issues
Fixes https://github.com/elastic/terraform-provider-ec/issues/471

## Motivation and Context
Currently, managing the autoscaling limits on currently unsized topology elements results in an endless diff loop. The provider filters out all unsized topology elements when storing the resource state, meaning the defined topology elements are seen as new elements on a subsequent apply.   

## How Has This Been Tested?

Unit tests, manually

## Types of Changes

- [ ] Documentation

Marking this as a breaking change, since deployments which have autoscaling enabled will now include *all* topology elements in the resource state. This may require changes to existing resource declarations which were previously only defining a subset of possible topology elements. 

## Readiness Checklist


AlexP-Elastic (approved):
LGTM, not super familiar with either how TF works or go, but I could read the logic and it made sense to me why it was wrong before and why it is now correct :)

Issue #401, opened 2021-10-18
Specifying autoscaling limits on tiers which do not support them results in an API error - JSON request does not comply with schema ([]: [Enum string expected])


maaydin:
Defining autoscaling attributes on an Elasticsearch topology element which does not support that attribute (e.g `min_size` on a data tier, any autoscaling attribute on a master tier) fails with:

```
JSON request does not comply with schema ([]: [Enum string expected])
```

The provider enriches the deployment template with information from the Terraform module definition. The provider relies on the deployment template specifying the resource type in the `autoscaling_min_size` and `autoscaling_max_size` attributes. Data tiers do not include the `autoscaling_min_size` attribute, similarly master tiers do not include either the `autoscaling_min_size` or `autoscaling_max_size`. This means that the final request includes a size properly without the corresponding resource, e.g:

```
autoscaling_min_size: {
  size: 2048
}
```

vs 

```
autoscaling_min_size: {
  size: 2048
  resource: memory
}
```

Possible fixes: 
* Error in the provider when a autoscaling dimension is configured in the provider but not supported in the template
* Fix the resource to `memory` (we assume as much already)

# Original description
------------------

## Expected Behavior
When a dedicated master topology added to an existing cluster, it's expected that new dedicated master nodes added to cluster.

## Current Behavior
When a dedicated master topology added to an existing cluster, it returns below api error. 
`* api error: root.invalid_json_request: JSON request does not comply with schema ([]: [Enum string expected])`

### Full terraform apply log
```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.main.module.elasticcloud["***************"].ec_deployment.elasticcloud will be updated in-place
  ~ resource "ec_deployment" "elasticcloud" {
        id                     = "******************"
        name                   = "**********************"
        tags                   = {}
        # (8 unchanged attributes hidden)


      ~ elasticsearch {
            # (7 unchanged attributes hidden)

          + topology {
              + id               = "master"
              + node_roles       = (known after apply)
              + node_type_master = "true"
              + size             = "2g"
              + size_resource    = "memory"
              + zone_count       = 3

              + autoscaling {
                  + max_size = "120g"
                }
            }

            # (2 unchanged blocks hidden)
        }

        # (2 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
[...]

PR #477, opened 2022-04-28
Provide a default for the autoscaling min/max resource

tobio:
## Description
This fixes an issue when specifying an autoscaling min/max when there is no corresponding min/max defined in the deployment template. 

We're unable to specify a default in the Terraform schema since the resource attributes are `Computed`. Instead we define a default in code. 

## Related Issues
Fixes https://github.com/elastic/terraform-provider-ec/issues/401

## How Has This Been Tested?
Unit tests

Manually with:

```
resource "ec_deployment" "staging" {
  region                 = "eu-west-1"
  deployment_template_id = "aws-storage-optimized"
  version = "7.17.3"

  elasticsearch {
    topology {
      id         = "hot_content"
      size       = "2g"
      zone_count = 3

      autoscaling {
        max_size = "8g"
      }
    }

    topology {
      id         = "warm"
      size       = "2g"
      zone_count = 2

      autoscaling {
        max_size = "15g"
        min_size = "4g"
      }
    }
  }
}
```

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/ec/ecresource/deploymentresource/elasticsearch_expanders.go
+++ b/ec/ecresource/deploymentresource/elasticsearch_expanders.go
@@ -184,6 +184,7 @@ func expandEsTopology(raw interface{}, topologies []*models.ElasticsearchCluster
 		if autoscalingRaw := topology["autoscaling"]; autoscalingRaw != nil {
 			for _, autoscaleRaw := range autoscalingRaw.([]interface{}) {
 				autoscale := autoscaleRaw.(map[string]interface{})
+
 				if elem.AutoscalingMax == nil {
 					elem.AutoscalingMax = new(models.TopologySize)
 				}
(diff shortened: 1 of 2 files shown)


ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -260,6 +239,33 @@ func expandEsTopology(raw interface{}, topologies []*models.ElasticsearchCluster
 	return res, nil
 }
 
+func expandAutoscalingDimension(autoscale map[string]interface{}, model *models.TopologySize, dimension string) error {
```
mieciu, line 242: I think it might be worth placing a short comment on why we need this method

mieciu (approved):
Minor nit about the comment, but that's definitely good to go as it is 👍

Issue #333, opened 2021-06-26
Can a deployment use itself for logging?


milesqi:
If do not create a dedicated deployment for logging, is it possible to use itself in the plan. example as below

```tf
resource "ec_deployment" "**example_observability**" {
  name = "my_example_deployment"

  region                 = "us-east-1"
  version                = data.ec_stack.latest.version
  deployment_template_id = "aws-io-optimized-v2"

  elasticsearch {}

  kibana {}

  observability {
    deployment_id = ec_deployment.**example_observability**.id
  }
}
```

marclop:
I believe that terraform resources cannot reference themselves. However, you could hardcode the deployment ID and set it as `observability.deployment_id = "DEPLOYMENT_ID"`.

It is important to note that, while setting up self-monitoring is allowed, **it is not recommended for production workloads** since it can easily cause your deployment to overload itself when the load > than the cluster allows it for and the observability data will only make things worse.

marclop:
I'll also add a little of color on how to go about and create a deployment with self monitoring. The deployment cannot be created with self monitoring enabled, but can be set afterwards, this is a known issue and might be fixed in the future.

milesqi:
that would be great if it gets fixed in future release. for development and testing deployments we do hope it just self-logged. I also found the work around to set it after deployment is created, however, it's not a good way to fit into our deployment pipeline. We will wait for the fix if it can come soon. thank you.

marclop:
@milesqi We currently have no estimated timeline on fixing this issue, I can keep you posted though.

scathatheworm:
I'm running into this problem when trying to configure a monitoring deployment, which has monitoring enabled for itself as well. Any plans for addressing this? Hardcoding the deployment id is really dirty and kinda defeats the purpose of infrastructure as code.

JoranDox:
Not sure if this should be a different issue, but it's very related so I thought I'd ping here first?

Basically I have 2 clusters, one for production, and one for testing, and they already exist so I know the id of the testing one. Then I gave the id of the testing one for both the logging destinations (as you don't want production to log itself to itself), but I get this:
![image](https://github.com/elastic/terraform-provider-ec/assets/7152733/85d46828-d726-4a6a-af68-64055d34176a)
The id is for the testing cluster, which for the testing cluster reverts to "self" instead, and causes the confusion.
I can fix this in the code of course, but it would be nice if those values were considered the same somehow.

gigerdo:
[...]

PR #502, opened 2022-06-16
Observability: Handle passing 'self' as target deployment-id.

gigerdo:
Passing 'self' as the deployment-id is considered the same as the actual deployment-id of the resource itself. Adjusts the diff-calculation to always map 'self' to the current deployment-id. When creating the deployment, the value stays as 'self' since the deployment-id will only be known after creation.

I updated the docs to show this new capability, but might need some improvement. It is now possible to use the following config
```
observability {
  deployment_id: "self"
}
```
To directly enable self-monitoring even when creating a deployment

## Related Issues

Fixes https://github.com/elastic/terraform-provider-ec/issues/333

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/docs/resources/ec_deployment.md
+++ b/docs/resources/ec_deployment.md
@@ -139,6 +139,13 @@ resource "ec_deployment" "example_observability" {
 }
 ```
 
+It is possible to enable observability without a second deployment, by keeping observability data in the current deployment. To enable this use `self` as `deployment_id`.
+```hcl
+observability {
+  deployment_id = "self"
+}
+```
+
 ### With Cross Cluster Search settings
 
 ```hcl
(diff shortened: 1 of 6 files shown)


mieciu:
A note to future-self when reviewing: we'd want this case (rewrite of `self` in response/deployment ID representation in statefile) to be covered with integration test 🙂

tobio (commented):
Can we add an acceptance test for this change?

docs/resources/ec_deployment.md
```diff
@@ -139,6 +139,13 @@ resource "ec_deployment" "example_observability" {
 }
 ```
 
+It is possible to enable observability without a second deployment, by keeping observability data in the current deployment. To enable this use `self` as `deployment_id`.
```
kilfoyle, line 142:
```suggestion
It is possible to enable observability without using a second deployment, by storing the observability data in the current deployment. To enable this, set `deployment_id` to `self`.
```
I might reword this a bit, but it's pretty clear as is so it's just a suggestion.

kilfoyle (approved):
LGTM for the text. 👍 
I added a small suggestion.

docs/resources/ec_deployment.md
gigerdo: Thanks, I like your suggestion better :+1:

tobio (approved):

PR #507, opened 2022-06-20
Elasticsearch configuration strategy support

kuisathaverat:
## Description

The current implementation uses `autodetect` configuration strategy, this strategy is fine in most cases. This PR allows to choose the configuration strategy from a set of pre-configured strategies.

## Motivation and Context

In our case we use SNAPSHOT versions and this strategy does not work as expected when we changed the Docker image for a new SNAPSHOT version, we need to perform a `rolling` with `groupBy` to `__all__` to perform a stop and start of all nodes at the same time to avoid version conflicts.

## How Has This Been Tested?

I have made a manual test and the PR includes some unit tests.

## Types of Changes

- [x] Documentation

## Readiness Checklist

### DIFF
--- a/docs/resources/ec_deployment.md
+++ b/docs/resources/ec_deployment.md
@@ -382,7 +382,7 @@ The optional `elasticsearch.trust_external` block, allows external trust relatio
 
 ##### Strategy
 
-The optional `elasticsearch.strategy` allows choosing the configuration strategy used to apply the changes. This is a fine tune setting, you do not change it unless you have a specific case where the `autodetect` do not cover your use case.
+The optional `elasticsearch.strategy` allows you to choose the configuration strategy used to apply the changes. You do not need to change this setting unless you have a specific case where the `autodetect` does not cover your use case.
 
 * `type` Set the type of configuration strategy [autodetect, grow_and_shrink, rolling_grow_and_shrink, rolling_all].
   * `autodetect` try to use the best associated with the type of change in the plan.
--- a/docs/resources/ec_deployment.md
+++ b/docs/resources/ec_deployment.md
@@ -387,8 +387,8 @@ The optional `elasticsearch.strategy` allow to choose the configuration strategy
 * `type` Set the type of configuration strategy [autodetect, grow_and_shrink, rolling_grow_and_shrink, rolling_all].
   * `autodetect` try to use the best associated with the type of change in the plan.
   * `grow_and_shrink` Add all nodes with the new changes before to stop any node.
-  * `rolling_grow_and_shrink` Add nodes one by one replacing the existing ones when the new node es ready.
-  * `rolling_all` Strop all nodes perform the changes and start all nodes.
+  * `rolling_grow_and_shrink` Add nodes one by one replacing the existing ones when the new node is ready.
+  * `rolling_all` Stop all nodes, perform the changes and start all nodes.
  
 #### Kibana
 
--- a/docs/resources/ec_deployment.md
+++ b/docs/resources/ec_deployment.md
@@ -207,6 +207,36 @@ resource "ec_deployment" "with_tags" {
 }
 ```
 
+### With configuration strategy
+
+```hcl
+data "ec_stack" "latest" {
+  version_regex = "latest"
+  region        = "us-east-1"
+}
+
+resource "ec_deployment" "with_tags" {
+  # Optional name.
+  name = "my_example_deployment"
+
+  # Mandatory fields
+  region                 = "us-east-1"
+  version                = data.ec_stack.latest.version
+  deployment_template_id = "aws-io-optimized-v2"
+
+  elasticsearch {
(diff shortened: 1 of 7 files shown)


docs/resources/ec_deployment.md
```diff
@@ -349,6 +380,16 @@ The optional `elasticsearch.trust_external` block, allows external trust relatio
 * `trust_all` (Optional) If true, all clusters in this external entity will be trusted and the `trust_allowlist` is ignored.
 * `trust_allowlist` (Optional) The list of clusters to trust. Only used when `trust_all` is `false`.
 
+##### Strategy
+
+The optional `elasticsearch.strategy` allow to choose the configuration strategy used to apply the changes.
```
tobio, line 385: IMO we should have something in here along the lines of "don't change this unless you know that you need to", it's not functionality customers should be concerned with.

ec/ecresource/deploymentresource/schema_elasticsearch.go
```diff
@@ -385,11 +387,10 @@ func newSnapshotSourceSettings() *schema.Schema {
 
 func newExtensionSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:        schema.TypeSet,
-		Set:         esExtensionHash,
+		Type:        schema.TypeList,
 		Description: "Optional Elasticsearch extensions such as custom bundles or plugins.",
 		Optional:    true,
-		MinItems:    1,
+		MaxItems:    1,
```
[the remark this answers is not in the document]
kuisathaverat: I will revert it, is a mistake change.

docs/resources/ec_deployment.md
kuisathaverat: I have added kind of a warning.

ec/ecresource/deploymentresource/elasticsearch_expanders.go
```diff
@@ -135,9 +144,44 @@ func expandEsResource(raw interface{}, res *models.ElasticsearchPayload) (*model
...
+		} else if rawValue == rollingGrowAndShrink {
+			strategy.RollingGrowAndShrink = new(models.RollingGrowShrinkStrategyConfig)
+		} else if rawValue == rollingAll {
+			strategy.Rolling = &models.RollingStrategyConfig{
+				GroupBy: "__all__",
+			}
+		}
+	}
+	return res
+}
```
tobio, line 183: IMO we should remove the return values entirely here. If we're going to keep returning something, then we should probably use the return value [here](https://github.com/elastic/terraform-provider-ec/pull/507/files#diff-ad781c304fc98e99326fac3c762954eda56bb8f7082fe00d7522bf6e43b8570dR154)

 ```suggestion
func expandStrategy(raw interface{}, strategy *models.PlanStrategy) {
	for _, rawStrategy := range raw.([]interface{}) {
		strategyCfg, ok := rawStrategy.(map[string]interface{})
		if !ok {
			continue
		}
		rawValue := strategyCfg["type"].(string)
		if rawValue == autodetect {
			strategy.Autodetect = new(models.AutodetectStrategyConfig)
		} else if rawValue == growAndShrink {
			strategy.GrowAndShrink = new(models.GrowShrinkStrategyConfig)
		} else if rawValue == rollingGrowAndShrink {
			strategy.RollingGrowAndShrink = new(models.RollingGrowShrinkStrategyConfig)
		} else if rawValue == rollingAll {
			strategy.Rolling = &models.RollingStrategyConfig{
				GroupBy: "__all__",
			}
		}
	}
}
```

**Review by tobio — COMMENTED** [review]
Code lgtm, sorry for the back and forth I thought about this a little more last night.

What's the behaviour here when the current plan strategy doesn't match the strategy in the TF plan? We shouldn't trigger any deployment changes as a result of a strategy change, we may need to define `DiffSuppressFunc` on the `strategyResourceSchema`.

File: docs/resources/ec_deployment.md [hunk 4]
```diff
@@ -349,6 +380,16 @@ The optional `elasticsearch.trust_external` block, allows external trust relatio
 * `trust_all` (Optional) If true, all clusters in this external entity will be trusted and the `trust_allowlist` is ignored.
 * `trust_allowlist` (Optional) The list of clusters to trust. Only used when `trust_all` is `false`.
 
+##### Strategy
+
+The optional `elasticsearch.strategy` allows choosing the configuration strategy used to apply the changes. This is a fine tune setting, you do not change it unless you have a specific case where the `autodetect` do not cover your use case.
```
alaudazzi, line 385:
```suggestion
The optional `elasticsearch.strategy` allows you to choose the configuration strategy used to apply the changes. You do not need to change this setting unless you have a specific case where the `autodetect` does not cover your use case.
```

alaudazzi (approved):
Left minor editing suggestions, otherwise LGTM.

tobio (approved):

PR #523, opened 2022-08-10
[docs] Guide how to add SAML to a deployment

Kushmaro:
## Description
Added a guide that should render on the terraform registry docs site, to explain how customers should approach spinning up an Elastic Cloud deployment with a preconfigured SAML provider.

## Motivation and Context
This is a common customer use case, that is a bit tricky to achieve.

## How Has This Been Tested?
Rendering wasn't tested, but outcome of article was.

## Types of Changes
- [x] Documentation


docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+```yaml
+xpack.security.authc.providers:
+    saml.auth0:
+        order: 1
+        realm: auth0
+        icon: logoSecurity
+        description: "Log in with Auth0"
+```
+
+And that's it! spinning up the above ec_deployment will created a deployment on Elastic Cloud, with a preconfigured name & an additional Auth0 SSO identity provider and login option, that's alreday configured when the deployment is spun up.
```
tobio, line 103:
```suggestion
And that's it! Spinning up the above ec_deployment resource will creat a deployment on Elastic Cloud, with a preconfigured name and an additional Auth0 SSO identity provider and login option, that's already configured when the deployment is spun up.
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+Let's take a closer look at one specific argument here:
+```terraform
+format("%s-%s",var.name,substr("${random_uuid.uuid.result}",0,6))
+```
+
+This will tell terraform to create a string, that looks like  ```<deployment-name>-<6-digits-of-uuid>``` 
+We'll configure the deployment alias field to be the same, so if my deployment is named "elastic-sso" it'll get created as:
+```elastic-sso-8f9f6s``` for example.
```
tobio, line 68:
```suggestion
`elastic-sso-8f9f6s` for example.
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+}
+
+```
+
+Let's take a closer look at one specific argument here:
+```terraform
+format("%s-%s",var.name,substr("${random_uuid.uuid.result}",0,6))
+```
+
+This will tell terraform to create a string, that looks like  ```<deployment-name>-<6-digits-of-uuid>```
```
tobio, line 66:
```suggestion
This will tell terraform to create a string, that looks like  `<deployment-name>-<6-digits-of-uuid>` 
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+        config{
+           user_settings_yaml = file("./kb.yml")
+        }
+    }
+}
+
+```
+
+Let's take a closer look at one specific argument here:
+```terraform
```
tobio, line 62:
```suggestion
```hcl
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+
+```terraform
+resource "random_uuid" "uuid" {}
+```
+
+This will allow us to use a randomally generated uuid in our deployment name, so that we can later use it within the alias and revieve a unique deployment name & url alias, even before our deploymemnt is actually created. 
+
+Creating the "ec_deployment" resource, will look like this:
+
+```terraform
```
tobio, line 26:
```suggestion
```hcl
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+
+A common use case for the Elastic Cloud Terraform provider i, is to spin up an elastic deployment preconfigured with an SSO Identity provider (SAML2.0 or OIDC based)
+
+Since such a configuration often results in a cyclic dependency, where the Elasticsearch cluster yaml file requires the cluster full url name, but the url has not been generated yet, we're going to use Elastic Cloud's alias feature.
+
+*Please note, this guide doesn't go through what's needed to be configured on the IDP side, as this can change for different identity providers*
+
+First, we'll create an "random_uuid" resource, like so:
+
+```terraform
```
tobio, line 18:
```suggestion
```hcl
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
+---
+subcategory: ""
+page_title: "Configuring an SSO provider for a deployment"
+description: |-
+    An example of how to spin up a deployment preconfigured with a SAML provider
+---
+
+# Configuring a SAML provider for an Elastic Cloud Deployment
+
+A common use case for the Elastic Cloud Terraform provider i, is to spin up an elastic deployment preconfigured with an SSO Identity provider (SAML2.0 or OIDC based)
```
tobio, line 10:
```suggestion
A common use case for the Elastic Cloud Terraform provider is to spin up an Elastic Cloud deployment preconfigured with an SSO Identity provider (SAML2.0 or OIDC based)
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+
+Since such a configuration often results in a cyclic dependency, where the Elasticsearch cluster yaml file requires the cluster full url name, but the url has not been generated yet, we're going to use Elastic Cloud's alias feature.
```
tobio, line 12:
```suggestion
Relying on the URL generated by Elastic Cloud creates a cyclic dependency, where the Elasticsearch cluster configuration requires the Kibana URL, which only exists after the deployment is created. We can avoid the cyclic dependency by configuring the deployment with an alias, allowing us to know the final Kibana URL before the deployment is created.
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
+---
+subcategory: ""
+page_title: "Configuring an SSO provider for a deployment"
+description: |-
+    An example of how to spin up a deployment preconfigured with a SAML provider
```
tobio, line 5:
```suggestion
    An example of how to create a deployment preconfigured with a SAML provider
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        topology {
+            id = "warm"
+            size = "8g"
+            zone_count = 2
+        }
+
+        config{
+           user_settings_yaml = templatefile("./es.yml",{kibana_url=format("https://%s-%s.kb.us-east-1.aws.found.io:9243",var.name,substr("${random_uuid.uuid.result}",0,6))})
```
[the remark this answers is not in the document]
tobio: AFAICT we can't define a single domain which covers all regions which IMO makes this guide pretty painful to follow. At a minimum, we should mention "first go and create a deployment in the target region and see the format of the Kibana url's in use". 

It would be nice if we could interchangeably use `found.io`, `elastic-cloud.com`, `cloud.es.io` etc within these cluster urls. It would be even nicer if a deployment alias was a separate resource which customers could create before creating a deployment and avoid this whole palaver in the first place.

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+
+*Please note, this guide doesn't go through what's needed to be configured on the IDP side, as this can change for different identity providers*
+
+First, we'll create an "random_uuid" resource, like so:
+
+```terraform
+resource "random_uuid" "uuid" {}
+```
+
+This will allow us to use a randomally generated uuid in our deployment name, so that we can later use it within the alias and revieve a unique deployment name & url alias, even before our deploymemnt is actually created.
```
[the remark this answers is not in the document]
Kushmaro: @tobio this isn't trivial to achieve though, which is why I'm using the uuid resource.
The problem here is a cyclic dependency between determining the value of the alias (or being able to generate well-known unique aliases *before* the deployment gets created) and setting it in the elasticsearch.yml configuration.

docs/guides/configuring-sso-ec-deployment.md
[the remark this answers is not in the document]
Kushmaro: i'll run it through fmt. I hardcoded these for sake of simplicity. this isn't about writing "correct code" its about showing how this can be achieved in the most simple form.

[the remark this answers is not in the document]
Kushmaro: @tobio it doesn't make the guide painful to follow, rather, Elastic Cloud hard to use. It's an ESS problem, that this guide won't ever solve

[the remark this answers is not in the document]
Kushmaro: I'll note the part that ESS has different URLs though just to make that bit clearer.. but I wouldn't want to introduce variables for it just for easier readability (as I'm not sharing the entire tf configuration, rather only the key parts to get this working)

docs/guides/configuring-sso-ec-deployment.md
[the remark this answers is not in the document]
tobio: Around 10% of this guide discusses generating and formatting a UUID so you can have a unique alias. Using a uuid is one way of having a unique alias, but it isn't the only way. IMO it's not a good default to be suggesting. 

IMO we're complicating the guide by adding more detail than is required to achieve the goal. We could simply say something along the lines of "we're going to set an alias, you'll have to make sure it's unique" rather than have to explain the `random_uuid` resource and resulting string formatting. I'd expect, for most customers `<org-name>-<deployment-name>` would be likely to be unique without the extra trailing noise.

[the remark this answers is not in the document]
Kushmaro: @tobio your assumption is correct only if you ever generate a single deployment.. which isn't the case for users who automate things with TF, for the most part.

Again, I want to "spoon feed" our customers and field, since I'm sure not "solving for alias uniqueness" in this guide, will just leave (mostly myself) with more questions about how to configure SAML on an EC Deployment...

docs/guides/configuring-sso-ec-deployment.md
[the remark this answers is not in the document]
tobio: > It's an ESS problem, that this guide won't ever solve

Is there a Cloud issue to improve that situation? 

> it doesn't make the guide painful to follow

1. Create a deployment in the Cloud console in the region you want to use
2. Take a note of the format of the cluster endpoints
3. Transpose that format into the kibana_url defined here

That sounds pretty painful to me, whether we include those steps in the guide or not in reality it's what we're asking our customers to do. To be clear, I don't expect this guide to fix the underlying issue, but surely it's an issue we'd like to fix in some way? Do you know if there's an existing issue to standardise the regional domains, or implement some other solution?

docs/guides/configuring-sso-ec-deployment.md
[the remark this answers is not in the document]
tobio: > which isn't the case for users who automate things with TF

It seems reasonable to expect they would have a different deployment name though. Either way, I don't think this is a hill to die on :). It's a valid approach, just maybe more verbose than I'd choose 👍

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+
+Let's take a closer look at one specific argument here:
+```terraform
+format("%s-%s",var.name,substr("${random_uuid.uuid.result}",0,6))
+```
+
+This will tell terraform to create a string, that looks like  ```<deployment-name>-<6-digits-of-uuid>``` 
+We'll configure the deployment alias field to be the same, so if my deployment is named "elastic-sso" it'll get created as:
```
alaudazzi, line 67:
```suggestion
You will configure the deployment alias field to be the same, so if the deployment is named "elastic-sso" it gets created as follows:
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+
+Then, using a variable in our es.yml file, and a terraform templating mechanism, we'll generate our proper es.yml file. Our variable is named kibana_url, as seen in the ec_deployment resource above: 
+
+```terraform
+    config{
+           user_settings_yaml = templatefile("./es.yml",{kibana_url=format("https://%s-%s.kb.us-east-1.aws.found.io:9243",var.name,substr("${random_uuid.uuid.result}",0,6))})
+        }
+```
+
+This specific template, will use our name and UUID to determine the url for our Elasticsearch deployment before its even created, and put it in our `es.yml` file.
```
alaudazzi, line 78:
```suggestion
This specific template uses your name and UUID to determine the URL for the Elasticsearch deployment before it is even created, and stores it in the `es.yml` file.
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+  idp.metadata.path: "https://<url-for-provider's metadata>"
+  idp.entity_id: "urn:myproject.us.auth0.com"
+  sp.entity_id:  "${kibana_url}/"
+  sp.acs: "${kibana_url}/api/security/saml/callback"
+  sp.logout: "${kibana_url}/logout"
+  attributes.principal: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn"
+  attributes.groups: "http://schemas.xmlsoap.org/claims/Group"
+```
+
+Our kibana file should also contain the SSO provider configuration for Kibana to display it as a login option:
```
alaudazzi, line 92:
```suggestion
The kibana file should also contain the SSO provider configuration for Kibana to display it as a login option:
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+```yaml
+xpack.security.authc.providers:
+    saml.auth0:
+        order: 1
+        realm: auth0
+        icon: logoSecurity
+        description: "Log in with Auth0"
+```
+
+And that's it! Spinning up the above ec_deployment resource will create a deployment on Elastic Cloud, with a preconfigured name and an additional Auth0 SSO identity provider and login option, that's already configured when the deployment is spun up.
```
alaudazzi, line 103:
```suggestion
And that's it! Spinning up the ec_deployment resource will create a deployment on Elastic Cloud, with a preconfigured name and an additional Auth0 SSO identity provider and login option that's already configured when the deployment is spun up.
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+Let's take a closer look at one specific argument here:
+```terraform
+format("%s-%s",var.name,substr("${random_uuid.uuid.result}",0,6))
+```
+
+This will tell terraform to create a string, that looks like  ```<deployment-name>-<6-digits-of-uuid>``` 
+We'll configure the deployment alias field to be the same, so if my deployment is named "elastic-sso" it'll get created as:
+```elastic-sso-8f9f6s``` for example.
+
+Then, using a variable in our es.yml file, and a terraform templating mechanism, we'll generate our proper es.yml file. Our variable is named kibana_url, as seen in the ec_deployment resource above:
```
alaudazzi, line 70:
```suggestion
Then, by using a variable in the `es.yml` file and a terraform templating mechanism, you can generate your own `es.yml` file. Your variable is named kibana_url, as seen in the ec_deployment resource: 
```

docs/guides/configuring-sso-ec-deployment.md
```diff
@@ -0,0 +1,103 @@
...
+page_title: "Configuring an SSO provider for a deployment"
+description: |-
+    An example of how to create a deployment preconfigured with a SAML provider
+---
+
+# Configuring a SAML provider for an Elastic Cloud Deployment
+
+A common use case for the Elastic Cloud Terraform provider i, is to spin up an elastic deployment preconfigured with an SSO Identity provider (SAML2.0 or OIDC based)
+
+Relying on the URL generated by Elastic Cloud creates a cyclic dependency, where the Elasticsearch cluster configuration requires the Kibana URL, which only exists after the deployment is created. We can avoid the cyclic dependency by configuring the deployment with an alias, allowing us to know the final Kibana URL before the deployment is created.
```
alaudazzi, line 12:
```suggestion
Relying on the URL generated by Elastic Cloud creates a cyclic dependency, where the Elasticsearch cluster configuration requires the Kibana URL, which only exists after the deployment is created. You can avoid the cyclic dependency by configuring the deployment with an alias, which allows you to know the final Kibana URL before the deployment is created.
```

alaudazzi (commented):
Left a few editing suggestions.
Avoid "above" as not accessible terminology.
Prefer `you` to `we/us` to make it more user oriented.

docs/guides/configuring-sso-ec-deployment.md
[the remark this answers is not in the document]
Kushmaro: @tobio it's an infra issue for the most part, i'd ask @yuvielastic

alaudazzi (approved):
LGTM

Kushmaro:
Since the jenkins failure doens't really seem to relate to this change (which is doc only), I'm merging despite the failure.

```
deployment_basic_test.go:44: Step 3/4 error: Error running apply: exit status 1
14:46:51          
14:46:51          Error: Provider produced inconsistent result after apply
14:46:51          
14:46:51          When applying changes to ec_deployment_traffic_filter.second, provider
14:46:51          "provider[\"registry.terraform.io/hashicorp/ec\"]" produced an unexpected new
14:46:51          value: Root resource was present, but now absent.
14:46:51          
14:46:51          This is a bug in the provider, which should be reported in the provider's own
14:46:51          issue tracker.
```

PR #539, opened 2022-09-26
fix: remove strategy DiffSuppressFunc

kuisathaverat:
## Description

We start using the strategy setting and we realized that the plan does not contain the setting in most of the cases, after checking what happens the cause is the DiffSuppressFunc, it removes the changes in most of the cases. Due to the setting being a transient setting and probably not extended used (probably us only) removing the DiffSuppressFunc does not have any impact.

## Related Issues

https://github.com/elastic/terraform-provider-ec/pull/507

## How Has This Been Tested?

* We create a cluster with the setting and we try to make some Docker image changes with the strategy setting configured to restart all nodes at the same time. Then we verify the plan does not have the proper configuration.
* Since then we are using a SNAPSHOT (with this change ) version of the Elastic Terraform provider to update our clusters and work as expected, the plan has the correct strategy configured.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation

## Readiness Checklist

- [X] My code follows the code style of this project
- [X] All new and existing tests passed

### DIFF
--- a/ec/ecresource/deploymentresource/schema_elasticsearch.go
+++ b/ec/ecresource/deploymentresource/schema_elasticsearch.go
@@ -538,10 +538,6 @@ func strategyResource() *schema.Resource {
 					}
 					return
 				},
-				// changes on this setting do not change the plan.
-				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return true
-				},
 			},
 		},
 	}


dimuon:
@kuisathaverat , can you please check why acceptance failed

tobio:
This test appears to be failing due to https://github.com/hashicorp/terraform-plugin-sdk/issues/1066 and https://github.com/hashicorp/terraform-plugin-sdk/issues/1073 (seem to be duplicates).

I've https://github.com/elastic/terraform-provider-ec/pull/542 the acceptance tests to run against 1.2.9 in CI until the TF issue is fixed.

tobio (approved):

kuisathaverat:
@dimuon The acceptance are passing , I did not change,  It looks like was not related to my PR, it have failed at `TestAccDeployment_withExtension` test

Issue #593, opened 2023-02-28
Error: failed retrieving the specified stack version -> Region [us-central1] not found


joelsdc:
## Readiness Checklist

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
Using the `data.ec_stack` with a GCP region should return the versions available in such region.

```hcl
data "ec_stack" "latest" {
  version_regex = "latest"
  region        = "us-central1"
}
```

## Current Behavior
```
│ Error: failed retrieving the specified stack version: 1 error occurred:
│ 	* api error: root.region_not_found: Region [us-central1] not found
```

## Steps to Reproduce
1. Define a `data.ec_stack`
2. Set region to `us-central1` (this is a GCP region)
3. Run terraform plan/apply.

## Context
Currently we cannot rely on the `ec_stack` data source, we need to manually control de version used in the `ec_deployment` resources.

## Possible Solution
Fix the provider so we can use GCP regions (my guess is that currently only AWS regions are supported).
I'm also guessing that this happens because currently there is no way to specify the cloud used and the provider is defaulting to AWS? This is guessing I haven't confirmed any of this.

## Your Environment
* Version used: Elastic Cloud Provider v0.5.1

joelsdc:
After doing a `ec_deployment` import I've noticed that it's using for region `gcp-us-central1`.

I tried using that in the `data.ec_stack` and it worked.

Can this be specified in the documentation?

https://registry.terraform.io/providers/elastic/ec/latest/docs/data-sources/ec_stack

```
* `region` (Required) - Region where the stack pack is. 
For Elastic Cloud Enterprise (ECE) installations, use "ece-region.
```

PR #597, opened 2023-03-06
Link through to the ESS region list in the ec_stack docs

tobio:
Fixes https://github.com/elastic/terraform-provider-ec/issues/593

Essentially copies the note from the `ec_deployment` resource docs, linking through to the list of available regions.


dimuon (approved):
👍

Issue #405, opened 2021-11-03
Add "ec_deployment_traffic_filter" data source


raynigon:
## Overview
In my current setup, i have multiple terraform projects.
In one project the basic infrastructure is setup (logging cluster, traffic filter, etc). 
This project will be refered to as `base` project.
In the other projects additional clusters are setup and the previously created traffic filter should be associated with these clusters.
These projects will be refered to as `customer`.
At the moment there is no way to get the id of the traffic filter which was created in the `base` project in one of the `customer` projects.
Therefore i have to copy the id manually to each project.
This could be prevented by adding a "ec_deployment_traffic_filter" data source,
which allows to query a traffic filter by name.

## Possible Implementation
Add `ec_deployment_traffic_filter` data source.

## Testing
Create a terraform project, with a `ec_deployment_traffic_filter` resource.
Add the `ec_deployment_traffic_filter` data source.
Compare if the output properties of the resource matches the properties of the data source.

## Context
Reuse a `ec_deployment_traffic_filter` in multiple terraform projects

## Your Environment
- One base project
- Multiple customer projects
- Shared `ec_deployment_traffic_filter` resource

PR #619, opened 2023-04-13
Add Traffic Filter data source

andrew-moldovan:
First contribution to this project so I'm approaching it slowly. The current implementation is super simple and just fetches the name when provided with an ID. (Also I'm sure there's a million things wrong with it still, never written Go before or touched the TF provider). 

The reason behind this PR is https://github.com/elastic/terraform-provider-ec/issues/405. The ask there is to be able to fetch the `id` based on the `name`, so reverse of what I have now. The problem is that I'm not sure we have an API that can do that? Anyone got any ideas?

/cc @dimuon @tobio 

## Types of Changes

- [ ] Documentation

## Readiness Checklist


tobio:
> The problem is that I'm not sure we have an API that can do that? Anyone got any ideas?

[List all rulesets](https://www.elastic.co/guide/en/cloud/current/Deployments_-_Traffic_Filter_-_CRUD.html#get-traffic-filter-rulesets) then do client side filtering on the name I think is the only solution at the moment.

ec/ecdatasource/trafficfilterdatasource/datasource.go
```diff
@@ -0,0 +1,113 @@
...
+	return tfsdk.Schema{
+		Attributes: map[string]tfsdk.Attribute{
+			"id": {
+				Type:        types.StringType,
+				Description: "The ID of the traffic filter.",
+				Required:    true,
+			},
+
+			// computed fields
+			"name": {
```
tobio, line 52: I'd expect that we'd have more of the traffic filter details available here as well. Is that the plan?

andrew-moldovan: Yep, was just keeping it simple for now. Though the original request just asks for the `id`. Do I really want to expose all of the properties?

tobio: I don't think we need to expose everything, I can see a use case where the traffic filter rules could be used to drive the configuration of firewall rules.

ec/ecdatasource/trafficfilterdatasource/datasource.go
```diff
@@ -0,0 +1,147 @@
...
+	client *api.API
+}
+
+var _ datasource.DataSource = &DataSource{}
+var _ datasource.DataSourceWithConfigure = &DataSource{}
+
+func (d *DataSource) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
+	return tfsdk.Schema{
+		Attributes: map[string]tfsdk.Attribute{
+			"name": {
```
tobio, line 45: There's a question about how flexible we want to make this. 

IMO we should support `name`, `region`, and `id` as optional filters. 

We could just return _all_ traffic filters and let users filter the list in a [for expression](https://developer.hashicorp.com/terraform/language/expressions/for#filtering-elements). Keeping it in the provider provides a clearer interface to the right data though.

ec/ecdatasource/trafficfilterdatasource/datasource.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+	}, nil
+}
+
+func rulesetsListSchema() tfsdk.Attribute {
+	return tfsdk.Attribute{
+		Description: "List of all Rulesets for this user.",
+		Computed:    true,
+		Attributes: tfsdk.ListNestedAttributes(map[string]tfsdk.Attribute{
+			"id": {
```
tobio, line 62: We definitely want the region here as well.

ec/ecdatasource/trafficfilterdatasource/datasource.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, ruleset := range res.Rulesets {
+		var m rulesetModelV0
+		m.Name = types.String{Value: *ruleset.Name}
+		m.Id = types.String{Value: *ruleset.ID}
+		var nameToFind types.String
+		nameToFind = types.String{Value: state.Name.Value}
+
+		if m.Name == nameToFind {
+			result = append(result, m)
+		}
```
tobio, line 143:
```suggestion
		if *ruleset.Name != state.Name.Value {
			continue
		}
		m := rulesetModelV0{
			Name: types.String{Value: *ruleset.Name},
			Id: types.String{Value: *ruleset.ID},
		}
		result = append(result, m)
```

dimuon:
@andrew-moldovan , I think it makes sense to add unit tests for `modelToState`, similar to `ec/ecresource/trafficfilterresource/flatteners_test.go` and an integration test for the new data source.

ec/ecdatasource/trafficfilterdatasource/datasource.go
andrew-moldovan: Yep agreed, I'm working on getting `rules` exposed but struggling with types and stuff. Making progress though

andrew-moldovan:
Yep, I will add tests still, and expose more of the attributes. Just struggling with Go and learning how this all works, so progress is a bit slow

andrew-moldovan:
@dimuon could you please take a look at the test I have tried to write? I am struggling with types, figuring out how they all go and why they're so awful 😂 I feel like what I have should _theoretically_ work, but it's yelling about something

andrew-moldovan:
Ok so I've got the unit tests up and running and working on the acceptance tests. Couple things:

1. Almost all of them seem to be failing and I have a hard time believing that adding an extra data source has made them all fail. They're all failing with `After applying this test step, the plan was not empty` and it looks like they might be trying to add `apm_secret_token`?
2. For my actual acceptance test, I basically want to make sure it can filter by id/name/region (or even just one of them) and return the number of rulesets I expect and some details of the ruleset. However, I'm not sure what account is being used and if this is checking our QA? or Staging? Can I add a single traffic filter that I can make sure won't be deleted? Right now there seems to be 42 rulesets for `us-east-1` so I'm worried that isn't likely to be a stable number. I see that the `datasource_stack` test essentially just checks to see if `version` is set but doesn't check to _what_ it's set. I guess I could use something similar, ie just make sure that _some_ rulesets are returned? But even that could be screwed if all the rulesets are deleted by some other process.

Any ideas @tobio @dimuon

dimuon:
@andrew-moldovan, the first issue should be already fixed in master branch.

dimuon:
@andrew-moldovan , can you leverage `ec_deployment_traffic_filter` to create a filter or two and then reuse them in acceptance test? E.g. `ec/acc/deployment_traffic_filter_test.go` creates such traffic filters.

andrew-moldovan:
@dimuon ah yes good call. I think I've got that working 🎉 

Marking this as ready to review now

examples/deployment/deployment.tf
```diff
@@ -19,6 +19,16 @@ data "ec_stack" "latest" {
...
+data "ec_trafficfilter" "test" {
+  # name = "example-filter"
+  # id = "41d275439f884ce89359039e53eac516"
+  region = "us-east-1"
+}
+
+output "test" {
+  value = data.ec_trafficfilter.test
+}
+
```
tobio, line 31:
```suggestion
```

I don't think this makes sense here?

tobio (commented):
We should add some [docs](https://github.com/elastic/terraform-provider-ec/tree/master/docs/data-sources) for this data source.

tobio (approved):
Small docs change, but lgtm. We should be generating these docs so happy to leave this as is for now and hopefully someo can automate it soon.

docs/data-sources/ec_trafficfilter.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+## Argument Reference
+
+* `name` (Optional) - The name of the traffic filter. Has to match exactly.
+* `id` (Optional) - The id of the traffic filter. Has to match exactly.
+* `region` (Optional) - Region where the traffic filter is. For Elastic Cloud Enterprise (ECE) installations, use `"ece-region`.
+
+## Attributes Reference
+See https://www.elastic.co/guide/en/cloud/current/definitions.html#TrafficFilterRulesets for the API guide
```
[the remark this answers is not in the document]
andrew-moldovan: I deliberately didn't document the actual structure here because I felt that wasn't the right choice. Why am I copy/pasting the structure from the link and writing up my own descriptions (or copying them) when that structure could change or new things added or descriptions changed? Then we'd have to remember to update them in these docs as well instead of just having people visit the one source of truth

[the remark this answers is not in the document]
andrew-moldovan: Just re-read your comment about autogenerating these docs and I totally agree with that!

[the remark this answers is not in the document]
tobio: > when that structure could change or new things added or descriptions changed

Because API structure changes aren't automatically reflected here. We'd have to add those fields to the provider as well.

Issue #614, opened 2023-03-31
Imported ec_deployment resource displays incorrect version


f1nzer:
## Readiness Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
`terraform state import` imports a correct version of ec_deployment

## Current Behavior

The version it imports is incorrect: it displays 8.4.3, instead of 8.7.0
![image](https://user-images.githubusercontent.com/1970236/229204552-569f43db-00b9-4e9f-998f-1e543b91cb34.png)

## Context

I had an ec_deployment of version 8.4.3.
I've updated my cluster manually to version 8.7.0 using elastic cloud dashboard.
Now I'm trying to import terraform state using latest ec provider (0.6.0).

I've tried to update it to 8.7.0 by changing my tf configuration (my expectation was that it will be a no-op):
```tf
resource "ec_deployment" "main" {
...
  version                = "8.7.0"
```

terraform plan partial output:
```
~ version                = "8.4.3" -> "8.7.0"
```

Unfortunately, `terraform apply` command throws exception:
```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to ec_deployment.main, provider
│ "provider[\"registry.terraform.io/elastic/ec\"]" produced an unexpected new
│ value: .version: was cty.StringVal("8.7.0"), but now
│ cty.StringVal("8.4.3").
│
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
```

## Your Environment

* Version used: 0.6.0
* Running against Elastic Cloud
* Terraform: 1.4.2
* Operating System and version: Windows 10

f1nzer:
Ok, I figured out what was the problem.

According to ElasticCloud API (/deployments/deploymentId) I had Enterprise Search of version 8.4.3, while all other resources are 8.7.0.

Many months ago, I had an Enterprise Search that was disabled (removed).
For some reason, it was not updated with other resources on the recent update to 8.7.0 - it is ok, because it was disabled/removed.

There is a part of response from /deployments endpoint:
```json
"enterprise_search": [
  "status": "stopped",
  "plan_info": {
    "current": {
        "plan": {
            "enterprise_search": {
                "version": "8.4.3"
            }
        }
    }
}
```

[...]

PR #623, opened 2023-04-17
Use the lowest version from a running resource, not just the lowest

tobio:
## Description

Updates the deployment version determination to filter out stopped resources. The API returns the current plan for resources which have at some point been running, which should not be considered when finding the lowest version.  

## Related Issues

Fixes https://github.com/elastic/terraform-provider-ec/issues/614

## How Has This Been Tested?

Updated an existing acceptance test to cover this scenario

## Types of Changes

- [ ] Documentation

## Readiness Checklist


ec/acc/deployment_post_node_role_upgrade_test.go
dimuon, line 1: @tobio , I suppose this test should fail without the fix (without the fix it should output `Error: Provider produced inconsistent result after apply` complaining about the version), correct?

dimuon (approved):
LGTM

dimuon:
@tobio , I think both unit and integration tests fail due to changes in core terraform, so if we use an older terraform cli (something like 1.3.9). the tests should pass. Anyway we need to address the issue ASAP but I believe we have also to move to the latest Framework version to handle that correctly.

tobio:
Yep, I fixed the TF version in #139

ec/acc/deployment_post_node_role_upgrade_test.go
tobio: Yep

Issue #621, opened 2023-04-14
Traffic filters in 0.6.0: "provider produced inconsistent result after apply"


jamieparkinson:
## Readiness Checklist
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
Planning of an `ec_deployment` with associated `ec_deployment_traffic_filter_association`s should be idempotent: after applying once no changes should remain in the plan, and subsequent applys should succeed regardless.

## Current Behavior
A cluster with associated traffic filters has successfully been configured/applied, but a subsequent plan aims to remove the traffic filters. On applying this plan, the following error occurs (specifics have been replaced with pseudocode in angle brackets):

```
Error: Provider produced inconsistent result after apply

When applying changes to <ec_deployment resource>, provider "provider[\"registry.terraform.io/elastic/ec\"]" produced an unexpected new value: .traffic_filter: was null, but now cty.SetVal([]cty.Value{cty.StringVal("<filter id 1>"), cty.StringVal("<filter id 2>")}).

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

We are also seeing spurious config changes to the `elasticsearch` block in the deployment resource, but I think this is already being tracked in other issues on this repo.

## Terraform resources
```hcl
resource "ec_deployment" "cluster" {
  // name, alias etc config omitted for brevity
  elasticsearch = {
    autoscale = "false"
    hot = {
      size        = "1g"
      zone_count  = 1
      autoscaling = {}
    }
  }
}

resource "ec_deployment_traffic_filter_association" "cluster" {
  for_each = var.network_config.ec_traffic_filters // set of filter ids

  deployment_id     = ec_deployment.cluster.id
  traffic_filter_id = each.value
}
```

## Steps to Reproduce

1. Apply a plan with `ec_deployment_traffic_filter_association`s
2. Plan again - see that filters are to be removed
3. Apply again - see error above.

## Your Environment
Using the 0.6.0 provider

jamieparkinson:
Related to https://github.com/elastic/terraform-provider-ec/issues/419

tobio:
This got released with 0.8.0 yesterday.

Issue #419, opened 2021-12-20
updates to ec_deployment resource destroys ec_deployment_traffic_filter_association resources


chrisjprior:
## Readiness Checklist
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
When applying an update to ec_deployment resource without the traffic_filter attribute set, terraform should not destroy related ec_deployment_traffic_filter_association resources.

From the documentation:

> Note on traffic filters
> If you use traffic_filter on an ec_deployment, Terraform will manage the full set of traffic rules for the deployment, and treat additional traffic filters as drift. For this reason, traffic_filter cannot be mixed with the ec_deployment_traffic_filter_association resource for a given deployment.

## Current Behavior
ec_deployment_traffic_filter_association resources linked to the ec_deployment resource which is being updated are destroyed. The provider is treating the ec_deployment resource as if it has its traffic_filter attribute set to an empty list.

## Terraform Definition
```
data "ec_stack" "stack" {
  version_regex = var.stack_version
  region        = var.region
}

resource "ec_deployment" "elasticsearch" {
  name = var.name

  region                 = data.ec_stack.stack.region
  version                = data.ec_stack.stack.version
  deployment_template_id = "aws-general-purpose" 

  elasticsearch {
    autoscale = "false"
    topology {
      id = "hot_content"
      size = "8g"
      zone_count = length(var.azs)
    }

    trust_account {
      account_id = data.external.account.result.id
      trust_all = false
      trust_allowlist = []
    }
  }

  kibana {
    topology {
      size = "1g"
      zone_count = length(var.azs)
    }
  }

 apm {
   topology {
     size = "1g"
   }
 }
 
  tags = {
    Name = var.name
  }
}

resource "ec_deployment_traffic_filter" "deny" {
  name   = "DENY"
  region = data.ec_stack.stack.region
  type   = "ip"

  rule {
    source = "127.0.0.1/32"
  }
}

resource "ec_deployment_traffic_filter_association" "deny" {
  traffic_filter_id = ec_deployment_traffic_filter.deny.id
  deployment_id     = ec_deployment.elasticsearch.id
}
```

## Steps to Reproduce
1. Create a project using the above terraform snippet and set the used variables
2. Run terraform apply (we are using provider v 0.3.0)
[...]

PR #632, opened 2023-04-30
Deployment traffic filters are computed to not conflict with a traffic filter association resource

tobio:
## Description

Marks the ec_deployment traffic_filter attribute as computed, since it can also be managed via a ec_deployment_traffic_filter_association resource. 

## Related Issues
Fixes https://github.com/elastic/terraform-provider-ec/issues/621
Fixes https://github.com/elastic/terraform-provider-ec/issues/419

## How Has This Been Tested?

Manually, acceptance tests 

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/ec/ecresource/deploymentresource/update.go
+++ b/ec/ecresource/deploymentresource/update.go
@@ -116,7 +116,7 @@ func (r *Resource) ResetElasticsearchPassword(deploymentID string, refID string)
 }
 
 func HandleTrafficFilterChange(ctx context.Context, client *api.API, plan, state v2.DeploymentTF) diag.Diagnostics {
-	if plan.TrafficFilter.IsNull() || plan.TrafficFilter.Equal(state.TrafficFilter) {
+	if plan.TrafficFilter.Equal(state.TrafficFilter) {
 		return nil
 	}
 
--- a/.changelog/632.txt
+++ b/.changelog/632.txt
@@ -1,0 +1,5 @@
+```release-note:bug-fix
+Prevents traffic filters managed with the `ec_deployment_traffic_filter_association` from being disassociated by the `ec_deployment` resource ([#419](https://github.com/elastic/terraform-provider-ec/issues/419)).
+This also fixes a provider crash for the above scenario present in 0.6 ([#621](https://github.com/elastic/terraform-provider-ec/issues/621))
+```
+https://github.com/elastic/terraform-provider-ec/pull/632
--- a/docs/resources/deployment.md
+++ b/docs/resources/deployment.md
@@ -268,7 +268,7 @@ resource "ec_deployment" "ccs" {
 - `request_id` (String) Request ID to set when you create the deployment. Use it only when previous attempts return an error and `request_id` is returned as part of the error.
 - `reset_elasticsearch_password` (Boolean) Explicitly resets the elasticsearch_password when true
 - `tags` (Map of String) Optional map of deployment tags
-- `traffic_filter` (Set of String) List of traffic filters rule identifiers that will be applied to the deployment.
+- `traffic_filter` (Set of String) List of traffic filters rule identifiers that will be applied to the deployment. Removing this attribute entirely *will not* remove managed traffic filters, instead first set it to an empty list (e.g `traffic_filter = []`) to remove the managed traffic filters.
 
 ### Read-Only
 
--- a/ec/acc/deployment_basic_test.go
+++ b/ec/acc/deployment_basic_test.go
@@ -35,10 +35,12 @@ func TestAccDeployment_basic_tf(t *testing.T) {
 	randomAlias := "alias" + acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 	trafficFilterCfg := "testdata/deployment_basic_with_traffic_filter_2.tf"
 	trafficFilterUpdateCfg := "testdata/deployment_basic_with_traffic_filter_3.tf"
+	emptyTrafficFilterCfg := "testdata/deployment_basic_with_empty_traffic_filter.tf"
 	resetPasswordCfg := "testdata/deployment_basic_reset_password.tf"
 	cfg := fixtureAccDeploymentResourceBasicWithAppsAlias(t, startCfg, randomAlias, randomName, getRegion(), defaultTemplate)
 	cfgWithTrafficFilter := fixtureAccDeploymentResourceBasicWithTF(t, trafficFilterCfg, randomName, getRegion(), defaultTemplate)
 	cfgWithTrafficFilterUpdate := fixtureAccDeploymentResourceBasicWithTF(t, trafficFilterUpdateCfg, randomName, getRegion(), defaultTemplate)
+	cfgWithEmptyTrafficFilter := fixtureAccDeploymentResourceBasicWithAppsAlias(t, emptyTrafficFilterCfg, randomAlias, randomName, getRegion(), defaultTemplate)
 	cfgResetPassword := fixtureAccDeploymentResourceBasicWithAppsAlias(t, resetPasswordCfg, randomAlias, randomName, getRegion(), defaultTemplate)
 	deploymentVersion, err := latestStackVersion()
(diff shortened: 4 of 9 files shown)


webfella (approved):
LGTM once the test is updated 👍

dimuon:
@tobio , I'm curious why the test fails. Did we come across a defect?

tobio:
> Did we come across a defect?

Previously the failing test was expecting that removing the `traffic_filter` attribute from a deployment should also remove the traffic filters from ESS. This PR breaks that, since it would then treat the attribute as Unknown and use the value from state in the plan, i.e leave the traffic filter present. 

It's a legitimate failure, I should push a fix for it shortly.

ec/ecresource/deploymentresource/traffic_filters.go
```diff
@@ -0,0 +1,95 @@
...
+func GetManagedFilters(ctx context.Context, req resource.UpdateRequest, privateData PrivateDataStore) ([]string, diag.Diagnostics) {
+	filterBytes, diags := privateData.GetKey(ctx, trafficFilterPrivateStateKey)
+	if diags.HasError() {
+		return []string{}, diags
+	}
+
+	var filters []string
+	if filterBytes == nil {
+		var filtersTF types.Set
+		diags = req.State.GetAttribute(ctx, path.Empty().AtName("traffic_filter"), &filters)
```
dimuon, line 81: diags = req.State.GetAttribute(ctx, path.Root("traffic_filter"), &filters)
```suggestion
		diags = req.State.GetAttribute(ctx, path.Empty().AtName("traffic_filter"), &filters)
```

dimuon, line 81:
```
diags = req.State.GetAttribute(ctx, path.Root("traffic_filter"), &filters)
```

ec/ecresource/deploymentresource/traffic_filters.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		diags.AddError("unable to prepare traffic filter private state", err.Error())
+		return diags
+	}
+
+	diags = privateData.SetKey(ctx, trafficFilterPrivateStateKey, data)
+	return diags
+}
+
+func CaptureManagedFiltersForUpdate(ctx context.Context, config tfsdk.Config, privateData PrivateDataStore) diag.Diagnostics {
```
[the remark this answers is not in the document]
tobio: Thanks, you're right. These were initially taking the request which is different. I'll fix these up.

ec/ecresource/deploymentresource/traffic_filters.go
dimuon: Actually it should be 
```
diags = req.State.GetAttribute(ctx, path.Root("traffic_filter"), &filtersTF)
```
I guess it can be executed if resource is imported instead of being created from scratch.

dimuon: Still, I don't see how the new implementation differs from the previous one for the failing test. The new implementation uses a value from config during creation, while the old implementation used state during update. The state during update should be the same as config during creation time if the field was set. 

I think we came across a TF defect for computed `set` - when a set field is computed, TF doesn't "see" a change for it. So the test fails because Update isn't called at all. 

Also, when `ec_deployment_traffic_filter` is deleted, is the rule for the deployment deleted as well? So basically `ec_deployment` shouldn't do anything at all - the rule is deleted when the appropriate `ec_deployment_traffic_filter` resource is deleted. The test fails due to the order of operations - it looks like `ec_deployment` resource is handled before `ec_deployment_traffic_filter` in the test.

ec/ecresource/deploymentresource/deployment/v2/remove_managed_traffic_filters_when_null.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"golang.org/x/exp/slices"
+)
+
+// Use current state for a topology's attribute if the topology's state is not nil and the template attribute has not changed
```
dimuon, line 28: The comment is not correct.

ec/ecresource/deploymentresource/deployment/v2/remove_managed_traffic_filters_when_null.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, filter := range managedFilters {
+		for i, elem := range planAsSet.Elems {
+			elemStr := elem.(types.String)
+			if elemStr.Value == filter {
+				planAsSet.Elems = slices.Delete(planAsSet.Elems, i, i+1)
+			}
+		}
+	}
+
+	if len(planAsSet.Elems) == 0 {
```
dimuon, line 57:
```suggestion
	if len(planAsSet.Elems) == 0 && !planAsSet.IsUnknown() {
```

dimuon:
@tobio , will it work in case of imported resource?

dimuon:
I'm curious whether the plan modifier can lead to the inconsistent result errors. E.g. once the modifier set new value to `traffic_filter` and subsequent read returns another value from backend, e.g. a rule that is added by traffic filter assoc resource or UI.
Can we just fix the test by explicitly setting `traffic_filter = []` instead of relying on the plan modifier?

tobio:
> I'm curious whether the plan modifier can lead to the inconsistent result errors. E.g. once the modifier set new value to traffic_filter and subsequent read returns another value from backend, e.g. a rule that is added by traffic filter assoc resource or UI.

This shouldn't be the case, and I've tested the flow where a rule is added with a `ec_deployment_traffic_filter` at the same time as all rules are removed from the `ec_deployment` resource (i.e the plan modifier actually modifies the plan). I'll double check the UI case, since that's theoretically different. 

> The test fails due to the order of operations - it looks like ec_deployment resource is handled before ec_deployment_traffic_filter in the test.

No, the test is setting the `ec_deployment.traffic_filter` attribute, not introducing a `ec_deployment_traffic_filter` resource. 
 
> Can we just fix the test by explicitly setting traffic_filter = [] instead of relying on the plan modifier?

Nope, I initially wanted to do this but basically it doesn't work. Explicitly setting this to an empty set led to a bunch of inconsistent result errors, to the point where the private state approach was simpler in the end. 

Setting this to an empty set required updating the read result to set state to an empty set when traffic filters are not included in the API response. There were edge cases around having an explicit empty set in the `ec_deployment` resource, and rules applied via `ec_deployment_traffic_filter` leading to inconsistent results as well. 

The private state approach theoretically allows users to manage traffic filters in the `ec_deployment` and `ec_deployment_traffic_filter` resources concurrently (I've tested this, it's works as expected). Whilst I wouldn't suggest that approach it does remove some hard edges from the provider.

dimuon (approved):

dimuon (approved):
👍

dimuon:
It looks like 2 acc tests related to traffic filters fail.

ec/ecresource/deploymentresource/update.go
```diff
@@ -90,7 +90,7 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 }
 
 func HandleTrafficFilterChange(ctx context.Context, client *api.API, plan, state v2.DeploymentTF) diag.Diagnostics {
-	if plan.TrafficFilter.IsNull() || plan.TrafficFilter.Equal(state.TrafficFilter) {
+	if plan.TrafficFilter.Equal(state.TrafficFilter) {
```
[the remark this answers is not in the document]
dimuon: I think that's the reason of the failure in ec/acc/deployment_traffic_filter_association_test.go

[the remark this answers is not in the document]
dimuon: Shall we delete rules only when the plan is set to empty array?

Issue #595, opened 2023-03-01
Private Connection data sources not available in v0.6.0


joelsdc:
## Readiness Checklist

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
The `ec_gcp_private_service_connect_endpoint` datasource to work

## Current Behavior
```
│ Error: Invalid data source
│
│   on service_elastic.tf line 7, in data "ec_gcp_private_service_connect_endpoint" "default":
│    7: data "ec_gcp_private_service_connect_endpoint" "default" {
│
│ The provider elastic/ec does not support data source "ec_gcp_private_service_connect_endpoint".
```

## Terraform definition:

As per documentation:

```hcl
data "ec_gcp_private_service_connect_endpoint" "default" {
  region = "us-central1"
}
```

## Steps to Reproduce
1. Add a `ec_gcp_private_service_connect_endpoint` datasource
2. Try to run terraform plan/apply

## Context
After upgrading to v0.6.0 of the provider the `ec_gcp_private_service_connect_endpoint` datasource no longer works.
On v0.5.1 `ec_gcp_private_service_connect_endpoint` works correctly.

## Your Environment
* Version used: v0.6.0
* Terraform cli: v1.3.9

joelsdc:
Looking at the diff between v0.5.1 and v0.6.0 I have the feeling that **all datasources** are broken...

joelsdc:
@tobio tagging you.

tobio:
@joelsdc sorry about the frustration here. tl;dr I'd stick with 0.5.1 until there's a follow up release here. 

It's expected that the `ec_gcp_private_service_connect_endpoint`, `ec_aws_privatelink_endpoint`, and `ec_azure_privatelink_endpoint` data sources aren't available in 0.6. The release notes indicate that changes made in 0.5x are not expected to be part of 0.6, however we can definitely be more explicit about what's missing. I'll chase this up. 

At some point during the development of the migration to the TF plugin framework (i.e the 0.6.0 release) it was decided to perform a wholesale migration based on the 0.4 branch, and follow up with the 0.5 changes in a later release. The result is a pretty frustrating user experience, I understand this is far from ideal for you right now. 

[...]

PR #659, opened 2023-05-30
Migrate private link datasources

tobio:
Fixes https://github.com/elastic/terraform-provider-ec/issues/595

Migrates the private link data sources to the plugin framework and re-adds them to the provider data sources.


ec/ecdatasource/privatelinkdatasource/datasource_test.go
```diff
@@ -0,0 +1,59 @@
...
+	t.Run("should error out when accessing an unknown provider", func(t *testing.T) {
+		source := privateLinkDataSource[v0AwsModel]{csp: "ibm"}
+		model := v0AwsModel{
+			RegionField: "us-north-7",
+		}
+
+		_, err := source.readRegionData(model)
+		require.ErrorIs(t, err, errUnknownProvider)
+	})
+	t.Run("should return a populate state model when accessing a valid region", func(t *testing.T) {
```
dimuon, line 45: Does it make sense to loop over all supported csp providers and regions?

dimuon:
I guess we have to update changelog as well.

dimuon (approved):
👍 nice generics usage. LGTM besides 2 nit-picks.

Issue #433, opened 2022-02-03
Race condition on creating Deployment with custom config that relies on keystore secret


jadrol:
In scenario when you specify custom `elasticsearch.yaml` settings using `elasticsearch.config.user_settings_yaml` and settings provided there requires you to use `keystore` for storing secure parameters you end in race condition when creating new deployment / editing existing.

## Readiness Checklist

- [x] I am running the latest version (*yes 3.0*)
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

Apply should work without failing.

## Current Behavior

Example of configuration documentation: https://www.elastic.co/guide/en/cloud/current/ec-securing-clusters-oidc-op.html#ec-securing-oidc-okta (steps 2 & 3)

What is happening:
1. Deployment is being created/updated with new `elasticsearch.yaml` settings
2. Keystore secret depends on deployment so it is waiting for 1. to finish
3. Elasticsearch cannot be restarted in step 1. which makes apply either hang (when creating deployment from scratch) or immediately fail (when applying to existing deployment).

## Terraform definition

```terraform
resource "ec_deployment" "race-condition-cluster" {
  name = "race-condition-cluster"

  region                 = "eu-west-1"
  version                = "7.16.3"
  deployment_template_id = "aws-observability"

  elasticsearch {
    topology {
      id         = "hot_content"
      size       = "1g"
      zone_count = 1
    }

    config {
      user_settings_yaml = <<EOF
xpack.security.authc.realms.oidc.oidc1:
  order: 2
  rp.client_id: client_id
  rp.response_type: "code"
  rp.requested_scopes: ["openid", "email"]
  rp.redirect_uri: "https://kibana.example.com/api/security/oidc/callback"
  op.issuer: "https://sample.okta.com"
  op.authorization_endpoint: "https://sample.okta.com/oauth2/v1/authorize"
  op.token_endpoint: "https://sample.okta.com/oauth2/v1/token"
  op.userinfo_endpoint: "https://sample.okta.com/oauth2/v1/userinfo"
  op.endsession_endpoint: "https://sample.okta.com/oauth2/v1/logout"
  op.jwkset_path: "https://sample.okta.com/oauth2/v1/keys"
  claims.principal: email
  claim_patterns.principal: "^([^@]+)@sample\\.com$"
EOF
    }
  }

  kibana {
    topology {
[...]

PR #674, opened 2023-07-05
Support managing keystore entries during deployment creation and update

dimuon:
## Description

Make use of the modified ECE Deployment API that enables specifying ES keystore contents during Create/Update calls.
It should make it possible to handle ES keystore secrets using deployment resource that got a new [attribute](https://github.com/elastic/terraform-provider-ec/blob/93593c31d81a93330fca4c1f72638376bdc8a57e/docs/resources/deployment.md#nestedatt--elasticsearch--keystore_contents) ([example](https://github.com/elastic/terraform-provider-ec/blob/93593c31d81a93330fca4c1f72638376bdc8a57e/examples/resources/ec_deployment/with-keystore/deployment.tf)).

**The appropriate API change is implemented in ECE 3.7.0, ESS ms-96.
The new IT test (`TestAccDeployment_keystore`) is expected to fail until ESS ms-96 release. I tested it against my local env.**

## Related Issues

Rel: #433

## Types of Changes

- [x] Documentation

## Readiness Checklist


ec/ecresource/deploymentresource/read.go
```diff
@@ -180,6 +181,33 @@ func (r *Resource) read(ctx context.Context, id string, state *deploymentv2.Depl
 		deployment.Elasticsearch.Strategy = baseElasticsearch.Strategy.ValueStringPointer()
 	}
 
+	if baseElasticsearch != nil && deployment.Elasticsearch != nil && !baseElasticsearch.KeystoreContents.IsNull() {
+		ds := baseElasticsearch.KeystoreContents.ElementsAs(ctx, &deployment.Elasticsearch.KeystoreContents, true)
+		diags.Append(ds...)
+
+		// response.Resources.Elasticsearch[0].Info.Metadata.Raw
```
tobio, line 188:
```suggestion
```

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
```diff
@@ -150,6 +150,32 @@ func ElasticsearchSchema() schema.Attribute {
 				Optional:    true,
 				Validators:  []validator.String{stringvalidator.OneOf(strategiesList...)},
 			},
+
+			"keystore_contents": keystoreContentsSchema(),
+		},
+	}
+}
+
+func keystoreContentsSchema() schema.Attribute {
```
tobio, line 159: So, we're only planning to manage keystore items which are added to the deployment. That is, we won't remove keystore items added outside of the deployment resource?

Just making sure I'm reading the rest of the code correctly. I agree the above is the right approach.

ec/ecresource/deploymentresource/read.go
```diff
@@ -180,6 +181,33 @@ func (r *Resource) read(ctx context.Context, id string, state *deploymentv2.Depl
 		deployment.Elasticsearch.Strategy = baseElasticsearch.Strategy.ValueStringPointer()
 	}
 
+	if baseElasticsearch != nil && deployment.Elasticsearch != nil && !baseElasticsearch.KeystoreContents.IsNull() {
```
tobio, line 184: Can we add a comment on this block so it's easier to understand the intent here. IIUC:

* Read TF state into `deployment.Elasticsearch.KeystoreContents`
* Get the current keystore from the API
* Any keystore items in TF state but not in the API response are removed

The above ensures that:
* New keystore items in TF config are added
* Keystore items not managed by TF are left alone
* Keystore items previously managed, but removed from config are removed

Is the above correct?

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
tobio: Basically just understanding the lack of `computed: true` here as well.

tobio (approved):
LGTM, some questions to make sure I'm on the same page. I think we can add some comments to the read logic to make this clearer too. 

We need to add a changelog entry here.

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
dimuon: Yeah, the `ec_deployment` resource should handle only keystore items that are declared in its attribute and shouldn't interfere with the items managed  by `ec_deployment_elasticsearch_keystore` resources. The new IT test should cover this use case.
Also the keystore API is write only so the provider cannot read secrets.

ec/ecresource/deploymentresource/read.go
dimuon: Thank you for the question - I'll add a comment.

Issue #599, opened 2023-03-08
Infinite ec_deployment.elasticsearch.config drift when using ec 0.6.0 and terraform 1.4.0


lra:
## Readiness Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

```
$ terraform plan
No changes. Your infrastructure matches the configuration.
```

It should not try to recreate the cluster if we don't have any `ec_deployment.elasticsearch.config` set.

## Current Behavior

With:
```
resource "ec_deployment" "REDACTED" {
  name                   = "REDACTED"
  region                 = "azure-westus2"
  version                = "7.16.3"
  deployment_template_id = "azure-io-optimized-v3"

  elasticsearch = {
    cold = {
      autoscaling = {}
    }

    config = {}

    coordinating = {
      autoscaling = {}
    }

    frozen = {
      autoscaling = {}
    }

    hot = {
      autoscaling = {}
    }

    master = {
      autoscaling = {}
    }

    ml = {
      autoscaling = {}
    }

    warm = {
      autoscaling = {}
    }
  }

  kibana = {}
}
```

I get this infinite diff:

```
$ terraform plan
Terraform will perform the following actions:

  # ec_deployment.REDACTED will be updated in-place
  ~ resource "ec_deployment" "REDACTED" {
      + apm_secret_token       = (sensitive value)
      ~ elasticsearch          = {
          ~ cloud_id       = "REDACTED" -> (known after apply)
          ~ config         = {
              + plugins = (known after apply)
            }
          ~ http_endpoint  = "REDACTED" -> (known after apply)
          ~ https_endpoint = "REDACTED" -> (known after apply)
          ~ resource_id    = "REDACTED" -> (known after apply)
            # (12 unchanged attributes hidden)
        }
        id                     = "REDACTED"
        name                   = "REDACTED"
        # (5 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
And with
```
    config = {
      plugins = []
    }
```
I get:
```
Terraform will perform the following actions:

  # ec_deployment.REDACTED will be updated in-place
  ~ resource "ec_deployment" "REDACTED" {
      + apm_secret_token       = (sensitive value)
      ~ elasticsearch          = {
          ~ cloud_id       = "REDACTED" -> (known after apply)
[...]

PR #677, opened 2023-07-14
Use state for unknown for computed attributes which won't change during plan application

tobio:
## Description
These values may potentially change throughout a deployment lifecycle, but those changes will be picked up during the refresh phase, updating state prior to computing the plan

These values, when unset, should be 'known' as they are provided in the API response

This PR does *not* result in completely clean plans due to how traffic filters currently work. I'm not sure there's a straightforward solution there, as the traffic_filter_association resource can lead to traffic filters changing during the course of a plan application. Simply using state for unknown on this attribute leads to TF crashing as a result of an unexpected new value after apply. 

I need to bump the TF version used in the acceptance tests here too. 

## Related Issues

Fixes https://github.com/elastic/terraform-provider-ec/issues/599

## How Has This Been Tested?

Manually, unit tests

## Types of Changes

- [ ] Documentation

## Readiness Checklist


ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_config.go
```diff
@@ -44,6 +44,7 @@ func readElasticsearchConfig(in *models.ElasticsearchConfiguration) (*Elasticsea
 		return &ElasticsearchConfig{}, nil
 	}
 
+	config.Plugins = []string{}
```
dimuon, line 47: It can lead to the inconsistency error if config is empty. Also it's not needed - `append` can append to `nil`.

dimuon: I mean if TF configuration doesn't have `config`.

jamesagarside:
Hopefully should address #370 too

jamesagarside:
Hopefully should address https://github.com/elastic/terraform-provider-elasticstack/issues/370

dimuon (approved):
It looks like acceptance tests failed due to the mentioned change to ES config's `plugins` initialization.

Besides that issue, LGTM 👍 . Thank you for the PR!

tobio:
@jamesagarside correct, I've been explicitly testing that issue locally with this change as well.

jamesagarside:
🥳🥳🥳

tobio:
Will need to explicitly set cloud_id to computed on name changes or when kibana is added/removed.

jamesagarside:
[Terraform Provider](https://trello.com/c/RZaS7f2n/14-terraform-provider)

gigerdo (approved):

PR #740, opened 2023-11-21
Setup the Release pipeline

nkammah:
## Description
Migrate the Jenkins [Release pipeline](https://github.com/elastic/terraform-provider-ec/blob/master/.ci/pipelines/release.Jenkinsfile) to Buildkite.

Notes:
* I moved all the logic to load the GPG keys into the pre-command hook to benefit from secrets redaction features and move away from the file based approach used in the `.ci/Makefile`
* With the above point, I had to rename an env. variable to `GPG_FINGERPRINT_SECRET` instead of `GPG_FINGERPRINT` so that Buildkite would always redact it and we don't risk leaking it (with a set -x or printenv). I don't think it would affect anyone since this is used only in the CI flow.
* The k8s agent used for the build is rather big (`cpu: "8" memory:"32G") but it seems needed as go-releaser is very hungry on cross-compiling all the targets in parallel - I had to iterate quite a bit to get to that combo.
* Related to the above, I re-ordered the goarch options to put `amd64` at the end which helped with getting the build to succeed as opposed to getting to resource exhaustion - i'm not too sure why.
* The `.ci/Makefile` should still work to run the flow as a developer instead of ci.

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/.buildkite/release.yml
+++ b/.buildkite/release.yml
@@ -3,5 +3,5 @@ steps:
     command: .buildkite/release.sh
     agents:
       image: "docker.io/library/golang:1.21"
-      cpu: "8" # The release flow needs that many or hits https://github.com/goreleaser/goreleaser/issues/304
-      memory: "16Gi"
+      cpu: "8"
+      memory: "4G"
--- a/.buildkite/release.yml
+++ b/.buildkite/release.yml
@@ -1,3 +1,7 @@
 steps:
-  - label: ":wave: Greetings"
-    command: "echo 'My release pipeline!'"
+  - label: ":rocket: Release"
+    command: .buildkite/release.sh
+    agents:
+      image: "docker.io/library/golang:1.21"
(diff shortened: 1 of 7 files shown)


dimuon (approved):
LGTM

.buildkite/release.yml
```diff
@@ -1,3 +1,7 @@
 steps:
-  - label: ":wave: Greetings"
-    command: "echo 'My release pipeline!'"
+  - label: ":rocket: Release"
+    command: .buildkite/release.sh
+    agents:
+      image: "docker.io/library/golang:1.21"
+      cpu: "8" # The release flow needs that many or hits https://github.com/goreleaser/goreleaser/issues/304
```
jamietanna, line 6: Is this the right issue number? `Add config for number of parallel builds #304`?

tobio (approved):

Issue #754, opened 2023-12-07
Add instance_config_version to all resources schema


claudia-correia:
## Overview

All topology elements now contain an `instance_configuration_version` field. We should add this field to the schema. Also, it is now possible to use different `instance_configuration_id`s from the ones defined in the deployment template. So, we should make the necessary changes so that it's possible to set the `instance_configuration_id` for all resources.

## Testing

We should test the following:
1. The `instance_configuration_id` and `instance_configuration_version` fields can be read and updated for all resources. 
2. It's possible to use an `instance_configuration_id` different from the one in the template.

## Context

We currently can't specify an `instance_configuration_version`, which forces us to use the `instance_configuration_version`s defined in the template. The same goes for the `instance_configuration_id`, which currently can't be different from the ones in the template, which blocks us from using different hardware types for specific resources/ES tiers.

PR #755, opened 2023-12-07
Add instance_config_version to all resources schema

claudia-correia:
## Description
1. Add the `instance_configuration_version` field to all resources schema.
2. Make the `instance_configuration_id` field writable for ES topology elements.
3. Remove the validation that forces the `instance_configuration_id` to be defined in the deployment template (for stateless resources)
4. Add unit tests 

## Related Issues

Rel: https://github.com/elastic/terraform-provider-ec/issues/754

## How Has This Been Tested?

Unit tests were added to test reading and updating the `instance_configuration_id` and `instance_configuration_version` fields for all resources.

## Types of Changes

- [ ] Documentation

## Readiness Checklist


ec/ecresource/deploymentresource/apm/v2/schema.go
```diff
@@ -112,6 +112,13 @@ func ApmSchema() schema.Attribute {
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"instance_configuration_version": schema.Int64Attribute{
+				Optional: true,
+				Computed: true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
```
tobio, line 119: So I think this (and the same in the other non-elasticsearch topology elements) may blow up when the `deployment.deployment_template_id` attribute is changed, since the config version may change as a result. 

The elasticsearch schema uses `UseTopologyStateForUnknown` instead which explicitly handles this situation (setting the `instance_configuration_version` to unknown if the top level template is changed). I think we'll want to do something similar here.

claudia-correia: This a great point, thanks for bringing it up! I made the change in https://github.com/elastic/terraform-provider-ec/pull/755/commits/2cc4ec49892415293c56286ae3f44f2ee616524c.

ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_topology.go
```diff
@@ -75,6 +77,14 @@ func (topology ElasticsearchTopologyTF) payload(ctx context.Context, topologyID
 		return diags
 	}
 
+	if topology.InstanceConfigurationId.ValueString() != "" {
+		topologyElem.InstanceConfigurationID = topology.InstanceConfigurationId.ValueString()
+	}
+
+	if topology.InstanceConfigurationVersion.ValueInt64() > 0 {
```
dimuon, line 84: We can use `!topology.InstanceConfigurationVersion.IsUnknown()` instead to check whether plan has a value for the attribute. Also, I guess this is the answer to your question from the ticket.

ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_topology.go
```diff
@@ -38,30 +38,32 @@ import (
...
-	ZoneCount               int                               `tfsdk:"zone_count"`
-	NodeTypeData            *string                           `tfsdk:"node_type_data"`
-	NodeTypeMaster          *string                           `tfsdk:"node_type_master"`
-	NodeTypeIngest          *string                           `tfsdk:"node_type_ingest"`
-	NodeTypeMl              *string                           `tfsdk:"node_type_ml"`
-	NodeRoles               []string                          `tfsdk:"node_roles"`
-	Autoscaling             *ElasticsearchTopologyAutoscaling `tfsdk:"autoscaling"`
+	id                           string
+	InstanceConfigurationId      *string                           `tfsdk:"instance_configuration_id"`
+	InstanceConfigurationVersion int                               `tfsdk:"instance_configuration_version"`
```
dimuon, line 57: Does it make sense to use `*int` instead if the attribute's value can be omitted and we'd like to differentiate between `0` and the absence of any value?

dimuon: Actually, I think `int` should work - the backend's model uses `int32` and the field is used only for reading responses from backend.

dimuon:
I still need to complete the review but I guess it makes sense to create/modify acceptance test that covers migration between DT versions using the new field.

ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_topology.go
claudia-correia: I updated this to `*int` to differentiate between 0 and the absence of any value.

.changelog/755.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+resource/deployment: Add instance_configuration_version field to all resources and allow to update the instance_configuration_id to a value not defined in the template.
```
dimuon, line 2: Does it make sense to mention other changes?

docs/resources/deployment.md
```diff
@@ -328,6 +329,8 @@ Required:
 
 Optional:
 
+- `instance_configuration_id` (String) Computed Instance Configuration ID of the topology element
```
dimuon, line 332: IMO it doesn't make sense to mention "Computed" - it should be clear from the schema. The code still contains similar passages but we should change that.

ec/ecresource/deploymentresource/apm/v2/apm_read.go
```diff
@@ -95,3 +99,13 @@ func IsApmStopped(res *models.ApmResourceInfo) bool {
 	return res == nil || res.Info == nil || res.Info.Status == nil ||
 		*res.Info.Status == "stopped"
 }
+
+func SetLatestInstanceConfigInfo(currentTopology *Apm, latestTopology *models.ApmTopologyElement) {
+	if currentTopology != nil && latestTopology != nil {
+		currentTopology.LatestInstanceConfigurationId = &latestTopology.InstanceConfigurationID
+		if latestTopology.InstanceConfigurationVersion != nil {
+			latestVersion := int(*latestTopology.InstanceConfigurationVersion)
+			currentTopology.LatestInstanceConfigurationVersion = &latestVersion
```
dimuon, line 108:
```suggestion
			currentTopology.LatestInstanceConfigurationVersion = ec.Int(int(*latestTopology.InstanceConfigurationVersion))
```
Actually you did something like that in other places. 
It could be that the suggestion cannot be used as is without importing "github.com/elastic/cloud-sdk-go/pkg/util/ec".

.changelog/755.txt
claudia-correia: Yes, it does 🙂 updated in https://github.com/elastic/terraform-provider-ec/pull/755/commits/989b44e01f8ec11092f2d992e533002d8fb1ae3c.

docs/resources/deployment.md
claudia-correia: Updated in https://github.com/elastic/terraform-provider-ec/pull/755/commits/7873fee0d571c5c6ecb8ce49326fbc4f31aeea21.

ec/ecresource/deploymentresource/apm/v2/apm_read.go
claudia-correia: You're totally right, thanks for the suggestion! Updated in https://github.com/elastic/terraform-provider-ec/pull/755/commits/4b95ac5b2f6dbb1600bb93644cf3335007cc6a3f.

dimuon:
@claudia-correia , some acceptance tests failed. Let's sync on Monday regarding this.

ec/internal/planmodifiers/use_state_unless_template_changed.go
```diff
@@ -20,42 +20,72 @@ package planmodifiers
...
-	resp.Diagnostics.Append(diags...)
+	// if the config is the unknown value, use the unknown value otherwise, interpolation gets messed up
+	if configValue.IsUnknown() {
+		return false, nil
+	}
+
+	templateChanged, d := AttributeChanged(ctx, path.Root("deployment_template_id"), plan, state)
+	diags.Append(d...)
+
+	var migrateToLatestHw bool
```
dimuon, line 74: It should be a TF type I think - the attribute's value can be `Unknown`.

ec/internal/planmodifiers/use_state_unless_template_changed.go
```diff
@@ -20,42 +20,72 @@ package planmodifiers
...
+	resp.Diagnostics.Append(diags...)
+	if useState {
+		resp.PlanValue = req.StateValue
 	}
+}
+
+func (m useStateForUnknownUnlessTemplateChanged) UseState(ctx context.Context, configValue attr.Value, plan tfsdk.Plan, state tfsdk.State, planValue attr.Value) (bool, diag.Diagnostics) {
+	var diags diag.Diagnostics
 
-	if req.StateValue.IsNull() {
```
dimuon, line 44: I think this caused the recent errors in acceptance tests.

ec/internal/planmodifiers/use_state_unless_template_changed.go
claudia-correia: I was assuming that `unknown` would read as `false`, which is the behaviour we want in case the field is not specified in the TF config 🤔 Does this make sense?

ec/internal/planmodifiers/use_state_unless_template_changed.go
claudia-correia: Hmm, yeah, that makes sense! I believe we only need this behaviour for the `instance_configuration_version`, because the remaining fields can't be `null`. Maybe I could move this condition to `PlanModifyInt64`, WDYT?

claudia-correia:
@dimuon Thanks for the details and for investigating the test failures! 🙇‍♀️ 
Your suggestion sounds good to me! However, I proposed an alternative [above](https://github.com/elastic/terraform-provider-ec/pull/755#discussion_r1452271309) which consists of using the `null` value only for `Int64` fields, which could _maybe_ be simpler than looking at the parent element. What do you think about this suggestion?

ec/internal/planmodifiers/use_state_unless_template_changed.go
dimuon: Yes, it should work too. Though I'd prefer to explicitly control (e.g. via an argument) allowing `null` in state instead of relying on the modifier type.

claudia-correia: Ohh, that sounds great! That way we could specifically set `instance_configuration_version` to be a "nullable" field, but leave all of the others (including `zone_count`, which is also `Int64`) as "non-nullable". I'll add those changes to the PR 🙂

ec/internal/planmodifiers/use_state_unless_template_changed.go
dimuon: Yes, it makes sense. But while it works here, we still rely on the TF framework implementation to read "unknown" state into a type that doesn't support this.

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
```diff
@@ -60,9 +60,11 @@ func (plan DeploymentTF) getBaseUpdatePayloads(ctx context.Context, client *api.
 		Kibana:             template.DeploymentTemplate.Resources.Kibana,
 	}
 
-	// If the deployment template has changed then we should use the template migration API
-	// to build the base update payloads
-	if newDtId != prevDtId && prevDtId != "" {
+	// If the deployment template has changed or MigrateToLatestHardware is true, we should use the template migration
+	// API to build the base update payloads
+	migrateToLatest := plan.MigrateToLatestHardware.ValueBool() || (newDtId != prevDtId && prevDtId != "")
```
dimuon, line 65: Does it make sense to check that migration is available (as the plan modifier does) in this new case (when `MigrateToLatestHardware` is set) before calling migration API?

ec/ecresource/deploymentresource/deployment/v2/schema.go
```diff
@@ -130,6 +130,10 @@ func DeploymentSchema() schema.Schema {
 				Description: "Explicitly resets the elasticsearch_password when true",
 				Optional:    true,
 			},
+			"migrate_to_latest_hardware": schema.BoolAttribute{
+				Description: "When true, updates deployment according to the latest deployment template values.",
```
dimuon, line 134: Shall we mention that if either IC or its version is set in config, migration doesn't happen even if the flag is set?

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
dimuon: Also, I think we can do some optimisations for this new case (only when DT isn't modified) - `Update` always follows `Read` so we can store `migrateUpdateRequest` in `ReadRequest.Private` in `Read` and then get it from `UpdateRequest.Private` in `Update` using `ec/ecresource/deploymentresource/private_state.go`. However, feel free to merge without implementing this now.

ec/ecresource/deploymentresource/deployment/v2/schema.go
claudia-correia: DOne in https://github.com/elastic/terraform-provider-ec/pull/755/commits/3af275c83d212d4d0094711db9109095e53fc7eb.

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
claudia-correia: Thanks for the suggestions 🙏 

> Does it make sense to check that migration is available

Done in https://github.com/elastic/terraform-provider-ec/pull/755/commits/fa8a4499d789a7752a83623d64fffcf0b87671b3.

> I think we can do some optimisations for this new case

Done in https://github.com/elastic/terraform-provider-ec/pull/755/commits/a23377791b701b46b33ccbce8ebdefaae2073b15.

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
```diff
@@ -60,48 +63,63 @@ func (plan DeploymentTF) getBaseUpdatePayloads(ctx context.Context, client *api.
...
+
+	// If the deployment template has changed or MigrateToLatestHardware is true, we should use the template migration
+	// API to build the base update payloads
+	migrateToLatest := plan.MigrateToLatestHardware.ValueBool() || templateChanged
+
+	// Template migration isn't available for deployments using node types
+	if migrateToLatest && !planHasNodeTypes {
+		// If the template has changed, we can't use the migrate request from private state.
+		// In this case, we fetch a new update request again from the template migration API
+		if migrateTemplateRequest == nil {
```
dimuon, line 82: I guess we cannot use `migrateTemplateRequest` if `templateChanged` is true, can we?

dimuon: Shall we change this to
```suggestion
		if migrateTemplateRequest == nil || templateChanged {
```

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
```diff
@@ -60,48 +63,63 @@ func (plan DeploymentTF) getBaseUpdatePayloads(ctx context.Context, client *api.
...
+	}
+
+	templateChanged := newDtId != prevDtId && prevDtId != ""
+
+	// If the deployment template has changed or MigrateToLatestHardware is true, we should use the template migration
+	// API to build the base update payloads
+	migrateToLatest := plan.MigrateToLatestHardware.ValueBool() || templateChanged
+
+	// Template migration isn't available for deployments using node types
+	if migrateToLatest && !planHasNodeTypes {
```
dimuon, line 79: Does it make sense to make a check similar to `CheckAvailableMigration` here if template hasn't changed? Otherwise we will be calling migrate API or using cached migration request for every `Update` call if `migrateToLatest` is set - do we want this?

ec/ecresource/deploymentresource/update.go
```diff
@@ -83,7 +92,11 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 	updatePrivateStateTrafficFilters(ctx, resp.Private, planRules)
 	resp.Diagnostics.Append(v2.HandleRemoteClusters(ctx, r.client, plan.Id.ValueString(), plan.Elasticsearch)...)
 
-	deployment, diags := r.read(ctx, plan.Id.ValueString(), &state, &plan, res.Resources, planRules)
+	var newMigrateTemplateRequest *deployments.MigrateDeploymentTemplateOK
+	deployment, diags := r.read(ctx, plan.Id.ValueString(), &state, &plan, res.Resources, planRules, newMigrateTemplateRequest)
```
dimuon, line 96: If DT has not been changed in plan there is no need to make another migrate API call, I guess.

ec/internal/planmodifiers/use_state_unless_template_changed.go
```diff
@@ -20,48 +20,144 @@ package planmodifiers
...
+		return false, nil
+	}
+
+	templateChanged, d := AttributeChanged(ctx, path.Root("deployment_template_id"), plan, state)
+	diags.Append(d...)
+
+	var migrateToLatestHw bool
+	plan.GetAttribute(ctx, path.Root("migrate_to_latest_hardware"), &migrateToLatestHw)
+
+	isMigrationAvailable, d := CheckAvailableMigration(ctx, plan, state, path.Root(m.resourceKind))
```
dimuon, line 85: nit-pick: if `migrate_to_latest_hardware` isn't set, there is no need to check that migration is available

ec/internal/planmodifiers/use_state_unless_template_changed.go
dimuon: We can return from func here, if template changed.

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
claudia-correia: I intended to do this, I added a comment about it and everything but then forgot to add the clause 😅 thanks for pointing it out! Added in https://github.com/elastic/terraform-provider-ec/pull/755/commits/a69026c9d9ed2acf9c0f3872779d6be75fb05bb6.
The strange thing is that the acceptance tests worked without it 🤔

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
claudia-correia: I think this shouldn't be a problem because at this point we already defined that the state will be used for the fields, right? So, it shouldn't make a difference whether we use the base payloads or the payloads from the migrate request (but, please correct me if I'm wrong, as I may be confused... 🤔).

ec/ecresource/deploymentresource/update.go
claudia-correia: It shouldn't make another request because now we only perform the request inside `read` [if the template has changed](https://github.com/elastic/terraform-provider-ec/pull/755/commits/a69026c9d9ed2acf9c0f3872779d6be75fb05bb6) (as you suggested above) 🙂

ec/internal/planmodifiers/use_state_unless_template_changed.go
claudia-correia: Great point! 😉 added in https://github.com/elastic/terraform-provider-ec/pull/755/commits/a69026c9d9ed2acf9c0f3872779d6be75fb05bb6.

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
dimuon: Do you mean the logic in the plan modifiers? Is so, than I think the answer is no. E.g. the plan modifier sets the plan value for `Apm` (e.g. IC) to its current state, so the attribute is omitted from plan output. Then `Update` functions creates payload based on data from migration request that can have different values from the current state, sends it to backend and calls `read` that reads the new `Apm`'s state. If the new state of `Apm` indeed differs from the previous (that was used by the plan modifiers before `Update`), TF reports an inconsistency error in the provider.
Anyway, I think we don't want to specify payload from the migrate request if there is no migration available.

ec/ecresource/deploymentresource/update.go
dimuon: Ahh, that's correct - my bad!

ec/ecresource/deploymentresource/update.go
```diff
@@ -83,7 +92,11 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
...
+
+	// Store migrate request in private state
+	updatePrivateStateMigrateTemplateRequest(ctx, resp.Private, newMigrateTemplateRequest)
```
dimuon, line 99: `newMigrateTemplateRequest` is always nil here. But, anyway we can remove the line - I guess we only need to store it to cache in `Read`.

ec/ecresource/deploymentresource/create.go
```diff
@@ -95,9 +96,13 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
 		filters = request.Settings.TrafficFilterSettings.Rulesets
 	}
 
-	deployment, diags := r.read(ctx, *res.ID, nil, &plan, res.Resources, filters)
+	var migrateTemplateRequest *deployments.MigrateDeploymentTemplateOK
+	deployment, diags := r.read(ctx, *res.ID, nil, &plan, res.Resources, filters, migrateTemplateRequest)
 	updatePrivateStateTrafficFilters(ctx, resp.Private, filters)
 
+	// Store migrate request in private state
+	updatePrivateStateMigrateTemplateRequest(ctx, resp.Private, migrateTemplateRequest)
```
dimuon, line 104: `migrateTemplateRequest` is alwyas `nil` here too. I think we can remove the line too.

ec/ecresource/deploymentresource/read.go
```diff
@@ -65,7 +65,11 @@ func (r *Resource) Read(ctx context.Context, request resource.ReadRequest, respo
 	}
 
 	// use state for the plan (there is no plan and config during Read) - otherwise we can get unempty plan output
-	newState, diags = r.read(ctx, curState.Id.ValueString(), &curState, nil, nil, privateFilters)
+	var migrateTemplateRequest *deployments.MigrateDeploymentTemplateOK
+	newState, diags = r.read(ctx, curState.Id.ValueString(), &curState, nil, nil, privateFilters, migrateTemplateRequest)
+
+	// Store migrate request in private state
+	updatePrivateStateMigrateTemplateRequest(ctx, response.Private, migrateTemplateRequest)
```
dimuon, line 72: Similar to other places - `migrateTemplateRequest` is `nil` here.

ec/ecresource/deploymentresource/private_state.go
```diff
@@ -57,3 +60,38 @@ func updatePrivateStateTrafficFilters(ctx context.Context, state PrivateState, f
 
 	return state.SetKey(ctx, trafficFilterStateKey, filterBytes)
 }
+
+func readPrivateStateMigrateTemplateRequest(ctx context.Context, state PrivateState) (*deployments.MigrateDeploymentTemplateOK, diag.Diagnostics) {
```
dimuon, line 64: It makes sesne to create unit tests for the read and write funcs.

ec/ecresource/deploymentresource/private_state.go
```diff
@@ -57,3 +60,38 @@ func updatePrivateStateTrafficFilters(ctx context.Context, state PrivateState, f
...
+	migrationUpdateRequestBytes, diags := state.GetKey(ctx, migrationUpdateRequestKey)
+	if migrationUpdateRequestBytes == nil || diags.HasError() {
+		return nil, diags
+	}
+
+	var migrationUpdateRequest *models.DeploymentUpdateRequest
+	err := migrationUpdateRequest.UnmarshalBinary(migrationUpdateRequestBytes)
```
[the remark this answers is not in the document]
dimuon:
```suggestion
	var migrationUpdateRequest models.DeploymentUpdateRequest
```

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
claudia-correia: I see, that makes sense! Thanks for the clarification 🙇‍♀️ 
Updated in https://github.com/elastic/terraform-provider-ec/pull/755/commits/35aaecb011f9d4105b8e65a10fd28ebd0473b570.

ec/ecresource/deploymentresource/update.go
claudia-correia: Addressed in https://github.com/elastic/terraform-provider-ec/pull/755/commits/4c70d0034521bc6e2be803e5472948612fb64f2c.

ec/ecresource/deploymentresource/create.go
claudia-correia: Addressed in https://github.com/elastic/terraform-provider-ec/pull/755/commits/4c70d0034521bc6e2be803e5472948612fb64f2c.

ec/ecresource/deploymentresource/read.go
claudia-correia: Addressed in https://github.com/elastic/terraform-provider-ec/pull/755/commits/4c70d0034521bc6e2be803e5472948612fb64f2c.

ec/ecresource/deploymentresource/private_state.go
claudia-correia: Added in https://github.com/elastic/terraform-provider-ec/pull/755/commits/039678ffa9d5d2f37b96826199e221b749053e07.

ec/internal/planmodifiers/use_state_unless_template_changed.go
claudia-correia: Thanks for the tip 🙏 I did a little refactor in https://github.com/elastic/terraform-provider-ec/pull/755/commits/7860b3b95a00c994967cd8a0e71b67508c262a15, based on this strategy.

ec/ecresource/deploymentresource/apm/v2/apm_payload.go
```diff
@@ -117,6 +119,47 @@ func ApmPayload(ctx context.Context, apmObj types.Object, updateResources *model
...
+
+	apmState, diags := objectToApm(ctx, state)
+	if diags.HasError() {
+		return false, diags
+	}
+
+	// We won't migrate this topology element if 'instance_configuration_id' or 'instance_configuration_version' are
+	// defined on the TF configuration. Otherwise, we may be setting an incorrect value for 'size', in case the
+	// template IC has different size increments
+	if !apmPlan.InstanceConfigurationId.IsUnknown() || !apmPlan.InstanceConfigurationVersion.IsUnknown() {
```
dimuon, line 150: Looks like acceptance tests failed because `apmPlan` is `nil` in the test. We need to add the check for `nil` for both `apmPlan` and `apmState`. The same for other components.

claudia-correia: Added in https://github.com/elastic/terraform-provider-ec/pull/755/commits/a0cc0fc589310cf024c05ea20354af560401baaa. Thanks!

ec/ecresource/deploymentresource/read.go
```diff
@@ -174,6 +178,31 @@ func (r *Resource) read(ctx context.Context, id string, state *deploymentv2.Depl
...
+	if !deployment.HasNodeTypes() {
+		// The MigrateDeploymentTemplate request can only be performed for deployments that use node roles.
+		// We'll skip this logic for deployments with node types.
+		migrateTemplateRequest, err := r.client.V1API.Deployments.MigrateDeploymentTemplate(
+			deployments.NewMigrateDeploymentTemplateParams().WithDeploymentID(deployment.Id).WithTemplateID(deployment.DeploymentTemplateId),
+			r.client.AuthWriter,
+		)
+
+		// Store migrate request in private state
+		if readResponse != nil {
```
dimuon, line 190: Shall we do this after the error check in lines 194-197?

dimuon: Or the idea is that it should cache `nil` in case of an error?

dimuon (approved):
Great work! 👍

ec/ecresource/deploymentresource/read.go
claudia-correia: Addressed in https://github.com/elastic/terraform-provider-ec/pull/755/commits/59a1c9a4e9bdebd7e6bd78656cb465b39550abb1.

claudia-correia:
Merging the PR as the acceptance test failure seems unrelated.

PR #798, opened 2024-03-18
Remove `changelog_generation.yml` workflow.

gigerdo:
I'm wondering if we should remove this workflow:
The workflow doesn't work anymore due to missing permissions. The workflow wants to merge changes directly into master which is not allowed anymore. So the flow just fails with every merge.

It seems easier to just run this command once before releasing anyway. Basically open a PR with the correct `CHANGELOG.md` and then release that commit after it has been merged.

What's your opinion @dimuon @tobio


dimuon:
@gigerdo , yes, it can be a valid option. However, does it make sense to follow that [use of change log generation](https://github.com/elastic/go-sysinfo/blob/main/.github/workflows/release.yml#L30) without committing into `master` branch? IMO our current release workflow looks more complicated (requires more manual steps) than it could be.

gigerdo:
Yeah could probably also be generated during the release. The one downside is that it isn't checked in then. So at some point somebody still has to check in the generated version. Though maybe we don't actually need the generated changelog to be checked in? And just generate it adhoc every time.

dimuon:
Yeah, I think we can skip the check-in.

gigerdo:
Hm though it isn't possible right now it seems. After every release we also have to manually add the new version to `CHANGELOG.md`. Also there is a post-release task to remove all `.changelog` entries as well.

So I would say we for now just manually update the changelog, add a new version, remove the old entries.

dimuon (approved):
LGTM

Issue #724, opened 2023-10-10
data resource - deployment templates


Kushmaro:
As a customer I should be able to query Elastic Cloud through terraform, for which deployment templates IDs are currently available for me to use.

dimuon:
The complexity is somewhere between 3 and 5 that depends also on what attributes we'd like to support (e.g. topology?).

gigerdo:
@Kushmaro I opened a PR now implementing this datasource:
https://github.com/elastic/terraform-provider-ec/pull/799

If you have any additional fields or data that you would like to see in the datasource, let me know.

PR #799, opened 2024-03-21
Add datasource for deployment-templates

gigerdo:
## Description

- Adds a new datasource for deployment-templates.
- By default the hidden deployment templates are filtered out. I think this is useful to show only the preferred templates similar to the UI
  - It is possible to get the full list of older templates by setting `show_hidden: true` 

The data for the datasource looks like this:

```
> data.ec_deployment_templates.test
{
  "region" = "aws-eu-west-1"
  "show_hidden" = false
  "stack_version" = "8.7.0"
  "templates" = tolist([
    {
      "description" = "Suitable for ingestion use cases with 1-4 days of data available for fast access and for search use cases with indexing and querying workloads. Provides the most CPU resources per unit of RAM."
      "hidden" = false
      "id" = "aws-cpu-optimized-arm"
      "min_stack_version" = "7.14.0"
      "name" = "CPU optimized (ARM)"
    },
   [...]
  ])
}
```

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/724

## How Has This Been Tested?

- Some manual testing in QA with a simple tf file containing just the datasource

## Types of Changes

- [ ] Documentation

## Readiness Checklist


ec/ecdatasource/deploymenttemplates/datasource.go
```diff
@@ -0,0 +1,194 @@
...
+
+var _ datasource.DataSource = &DataSource{}
+var _ datasource.DataSourceWithConfigure = &DataSource{}
+
+func (d *DataSource) Metadata(ctx context.Context, request datasource.MetadataRequest, response *datasource.MetadataResponse) {
+	response.TypeName = request.ProviderTypeName + "_deployment_templates"
+}
+
+func (d *DataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
```
dimuon, line 46: IMO it makes sense to allow getting templates by ID as well.

ec/ecdatasource/deploymenttemplates/datasource.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+var _ datasource.DataSource = &DataSource{}
+var _ datasource.DataSourceWithConfigure = &DataSource{}
+
+func (d *DataSource) Metadata(ctx context.Context, request datasource.MetadataRequest, response *datasource.MetadataResponse) {
+	response.TypeName = request.ProviderTypeName + "_deployment_templates"
+}
+
+func (d *DataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
dimuon, line 45: nit-pick: it makes sense to move schema and `mapResponseToModel` to dedicated files.

dimuon (approved):
LGTM, but IMO it makes sense to allow getting templates by IDs.

Also, maybe it makes sense to return also other template fields, e.g. topology - IMO that data can be useful as input in deployment resources. @Kushmaro, WDYT? (but it can be also a next feature)

ec/ecdatasource/deploymenttemplates/datasource.go
dimuon: Just filtering by id in this datasource.

gigerdo:
Yeah, the topology could be useful to see the available instance-sizes 🤔

Kushmaro:
I agree with @dimuon that it makes sense to have the ability to get a template by id too (as I understand, this gets the template by name?)

not sure if the ID part *has* to be part of this PR, or can be left to a later stage though.

gigerdo:
> as I understand, this gets the template by name

Currently it lists templates in a region. The idea behind that being that as a user you wouldn't actually know the deployment-templates or their ids, so you can list all available ones.

When getting by id, what kind of information would you want to see in the template? (Since currently the main useful info is the id itself to be used when creating a deployment). 

Though it could be very useful to add the topology in the same format as used for the resource. Then it would be easier to configure a deployment by taking info from the template.

Kushmaro:
Ah gotcha @gigerdo . Yes, topology and Id are definitely the key aspects here, getting one by ID should be useful that way too

gigerdo:
Also I'm wondering if I should change `hidden` to `deprecated` as it is done here: https://www.elastic.co/guide/en/cloud/current/ec-regions-templates-instances.html

Although in our API we use `hidden`, I think it is more clear to call them deprecated to make it clear that the non-deprecated ones should be used.

gigerdo:
I updated the PR:
- Now it is also possible to filter by id
- The returned template contains topology information as well

docs/data-sources/deployment_templates.md
```diff
@@ -0,0 +1,315 @@
...
+  region = "aws-eu-west-1"
+}
+
+resource "ec_deployment" "my_deployment" {
+  name                   = "My Deployment"
+  version                = "8.12.2"
+  region                 = data.ec_deployment_templates.all_templates.region
+  deployment_template_id = data.ec_deployment_templates.all_templates.templates.0.id
+
+  elasticsearch = data.ec_deployment_templates.all_templates.templates.0.elasticsearch
```
dimuon, line 34: I'm curious how `node_roles` and `node_types` are handled in this case? IIRC, templates define both `node_types` and `node_roles` but I guess we cannot use both in deployment payload.

ec/ecdatasource/deploymenttemplates/read.go
```diff
@@ -0,0 +1,391 @@
...
+			fmt.Sprintf("Failed retrieving deployment template list: %s", err),
+		)
+		return
+	}
+
+	showDeprecated := data.ShowDeprecated.ValueBool()
+	filerById := data.Id.ValueString()
+
+	templates := mapResponseToModel(res, showDeprecated, filerById)
+	data.Templates = templates
```
dimuon, line 63:
```suggestion
	data.Templates = mapResponseToModel(res, showDeprecated, filerById)
```

docs/data-sources/deployment_templates.md
dimuon: Never mind :)
The datasource doesn't return either `node_roles` or `node_types`.

ec/ecdatasource/deploymenttemplates/read.go
```diff
@@ -0,0 +1,391 @@
...
+	var availableSizes []string
+	ic, found := icMap[element.InstanceConfigurationID]
+	if found {
+		if ic.DiscreteSizes != nil {
+			availableSizes = make([]string, 0)
+			for _, size := range ic.DiscreteSizes.Sizes {
+				availableSizes = append(availableSizes, util.MemoryToState(size))
+			}
+		}
+	}
```
dimuon, line 235: Can we move it to a function that can be reused across all stateless apps?

dimuon: I mean lines 226-235 (getting `availableSizes`)

ec/ecdatasource/deploymenttemplates/schema.go
```diff
@@ -0,0 +1,173 @@
...
+		Attributes: map[string]schema.Attribute{
+			"region": schema.StringAttribute{
+				Description: "Region to select. For Elastic Cloud Enterprise (ECE) installations, use `ece-region`.",
+				Required:    true,
+			},
+			"id": schema.StringAttribute{
+				Description: "Filters for a deployment template with this id.",
+				Optional:    true,
+			},
+			"stack_version": schema.StringAttribute{
```
dimuon, line 39: Does it make sense to allow using `stack_version` or `show_deprecated` if `id` is specified?

docs/data-sources/deployment_templates.md
gigerdo: It works because the template currently doesn't contain either field.

But actually I might remove this example since it is really not that useful. Since the code anyway uses the template as the default and when using this pattern you can't override any values, which pretty much everyone will want to do.

ec/ecdatasource/deploymenttemplates/schema.go
```diff
@@ -0,0 +1,173 @@
...
+			"max_size": schema.StringAttribute{
+				Computed: true,
+			},
+			"min_size_resource": schema.StringAttribute{
+				Computed: true,
+			},
+			"min_size": schema.StringAttribute{
+				Computed: true,
+			},
+		},
```
dimuon, line 146: Does it make sense to include tier autoscaling override too?

dimuon (approved):
LGTM 👍

dimuon:
@gigerdo , can you also please check why the acc test failed?

ec/ecdatasource/deploymenttemplates/schema.go
gigerdo: Which one do you mean? This `autoscalingSchema` is currently set for every tier/topology-element

dimuon: https://github.com/elastic/cloud-sdk-go/blob/master/pkg/models/elasticsearch_cluster_topology_element.go#L51

ec/ecdatasource/deploymenttemplates/schema.go
gigerdo: No, best to add a validator that id makes no sense together with the other two options.

ec/ecdatasource/deploymenttemplates/schema.go
gigerdo: I added them now.

ec/ecdatasource/deploymenttemplates/model_elasticsearch.go
```diff
@@ -18,13 +18,14 @@
...
 
 type elasticsearchModel struct {
-	HotTier          *topologyModel `tfsdk:"hot"`
-	CoordinatingTier *topologyModel `tfsdk:"coordinating"`
-	MasterTier       *topologyModel `tfsdk:"master"`
-	WarmTier         *topologyModel `tfsdk:"warm"`
-	ColdTier         *topologyModel `tfsdk:"cold"`
-	FrozenTier       *topologyModel `tfsdk:"frozen"`
-	MlTier           *topologyModel `tfsdk:"ml"`
+	AutoscalingEnabled *bool          `tfsdk:"autoscaling_enabled"`
```
dimuon, line 21: I think the corresponding `ec_deployement`'s attribute is called `autoscale`.
IMO it makes sense to use the same naming.

gigerdo: Yeah true

dimuon (approved):

Kushmaro:
@gigerdo @dimuon , can we please release a version of the provider after this one is merged?

PR #800, opened 2024-03-21
Update CODEOWNERS with new team name

gigerdo:
The team was renamed to `control-plane-stateful-applications`. This way the right people are added as reviewers.

@tobio If you want I can also add you here so you are still added as reviewer to new PRs. (But I can also remove you if the notifications are annoying).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # These owners will be the default owners for everything in
 # the repository.
-*       @elastic/cloud-applications-solutions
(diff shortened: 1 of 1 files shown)


dimuon (approved):
LGTM

tobio (approved):
Lgtm, I'm happy with whatever y'all think is best.

gigerdo:
I guess I'll leave you then so we can still get your insights :)

gigerdo:
Actually I just realized that somehow in this repo you are still in the applications team. So we don't need to add you individually.

PR #806, opened 2024-04-04
Fix acceptance pipeline cleanup.

gigerdo:
The step currently has a key of "" so the `pre-exit` code never runs (Referenced in `.buildkite/hooks/pre-exit`). This leads to old acceptance-test deployments piling up over time.

With this change the step should run after the tests and clean up hanging deployments.


dimuon (approved):
👍

PR #807, opened 2024-04-04
Update VERSION to 0.11.0

gigerdo:
- Update the version to 0.11.0 in makefiles and examples.
- Reset the changelog for the new version


.changelog/683.txt
dimuon, line 1: Do we want to remove the original changelog entries?

gigerdo: We decided to keep the changelog entries instead of deleting them after every release.

dimuon (approved):

Issue #635, opened 2023-05-03
Automatically add/remove master nodes from plan based on deployment size


arjanelolong2022:
## Resource definition

```
{    
    version: '8.6.1',
    deploymentTemplateId: 'aws-storage-optimized-v3',
    elasticsearch: {
      autoscale: 'true',
      topologies: [
        {
          id: 'hot_content',
          size: '8g',
          zoneCount: 2,
          autoscaling: {
            maxSize: '15g',
            maxSizeResource: 'memory'
          },
        },
        {
          id: 'warm',
          size: '0g',
          zoneCount: 1
        }
      ],
    }
}
```

One `tie breaker master` was automatically provisioned.

## Expected Behavior
Changes will still be applied even without a master node definition.

## Current Behavior
On apply the error below is thrown. 
```
* api error: clusters.cluster_invalid_plan: Cluster must contain at least a master topology element and a data topology element. 'master' node type is missing,'master' node type exists in more than one topology element (resources.elasticsearch[0].cluster_topology)
* api error: clusters.cluster_invalid_plan: Instance configuration [aws.es.master.c5d] does not allow usage of node types [data,ingest]. You must either change instance configuration or use only allowed node types [master]. (resources.elasticsearch[0].cluster_topology[4].instance_configuration_id)
```

We are also not able to define/add a master node since we have less than 6 data nodes.

## Context
After we have provisioned our deployment, we are no longer able to apply updates to our deployment. 

## Your Environment
We provision via Pulumi.

* Version used: v0.5.1

## Might be related

https://github.com/elastic/terraform-provider-ec/issues/343

pascal-hofmann:
I tested around with v 0.7.0 and different node counts and found out the following:
1. Starting from 6 nodes, you need to specify at least one master via terraform.
2. Below 6 nodes, you must not specify any masters via terraform.
3. When switching from 5 nodes to 6 nodes with master, terraform apply will fail (a subsequent plan will show no drift related to this change though):
```
| When applying changes to module.[deployment_1.ec](http://deployment_1.ec/)_deployment.this, provider "provider[\"[registry.terraform.io/elastic/ec](http://registry.terraform.io/elastic/ec)\"]" produced an unexpected new value: .elasticsearch.hot.node_roles: planned set element cty.StringVal("master") does not correlate with any element in actual.
```
4. Switching back from 6 nodes and master to only 5 nodes is not possible at all via terraform (it does work via UI though):
```
[...]

PR #814, opened 2024-04-24
Automatically add and remove dedicated masters.

gigerdo:
## Description

If the terraform config defines no master, the provider will add/remove it as needed:
- Added if the number of nodes is >= the dedicated master threshold
- Removed if the number of nodes is < the dedicated master threshold
- The master tier can still be explicitly set (For example to change the size of the dedicated master tier)

Regarding the implementation:
- Uses a `ModifyPlan` modifier to first calculate the number of nodes, then adds or removes the master tier
  - The main advantage of this is that the plan before applying already shows that a master is added/removed  
- It is necessary to load the template to have some info about default sizes etc.

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/635
(Most of the problems described in that issue have already been solved, this PR just adds automatically adding/removing master tier)

## Types of Changes

- [ ] Documentation

## Readiness Checklist


ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	if !config.MasterTier.IsNull() && !config.MasterTier.IsUnknown() {
+		// Master tier is explicitly configured -> No changes will be made
+		tflog.Debug(ctx, "UpdateDedicatedMasterTier: Could not enable master tier, as it has no instance-config.")
+		return
+	}
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
tobio, line 71:
```suggestion
```

I don't think we're adding any diagnostics since the last check here.

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,189 @@
...
+
+	var config es.ElasticsearchTF
+	resp.Diagnostics.Append(req.Config.GetAttribute(ctx, path.Root("elasticsearch"), &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !config.MasterTier.IsNull() && !config.MasterTier.IsUnknown() {
+		// Master tier is explicitly configured -> No changes will be made
+		tflog.Debug(ctx, "UpdateDedicatedMasterTier: Could not enable master tier, as it has no instance-config.")
```
tobio, line 66: This message doesn't correlate with the condition above, can it be updated to be clearer>

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,189 @@
...
+		return
+	}
+
+	var config es.ElasticsearchTF
+	resp.Diagnostics.Append(req.Config.GetAttribute(ctx, path.Root("elasticsearch"), &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !config.MasterTier.IsNull() && !config.MasterTier.IsUnknown() {
```
tobio, line 64: Should this check be moved to the top of this function? It feels like an opportunity to exit early before counting cluster nodes etc.

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,189 @@
...
+	if template.DeploymentTemplate == nil ||
+		template.DeploymentTemplate.Resources == nil ||
+		len(template.DeploymentTemplate.Resources.Elasticsearch) == 0 ||
+		template.DeploymentTemplate.Resources.Elasticsearch[0].Plan == nil {
+		return nil
+	}
+	var masterTier *models.ElasticsearchClusterTopologyElement
+	for _, topology := range template.DeploymentTemplate.Resources.Elasticsearch[0].Plan.ClusterTopology {
+		if topology.ID == "master" {
+			masterTier = topology
```
tobio, line 121: Could break out of the loop here.

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,189 @@
...
+		size, err := deploymentsize.ParseGb(topology.Size.ValueString())
+		if err != nil {
+			tflog.Debug(ctx, "countNodesInCluster: Failed to parse topology size.", map[string]interface{}{"error": err})
+			continue
+		}
+
+		zoneCount := topology.ZoneCount.ValueInt64()
+		if size > 0 && zoneCount > 0 {
+			nodesInDeployment += zoneCount
+		}
```
tobio, line 185: How does this handle the case when `size` is greater than the maximum node size? That is there are multiple nodes for a given tier within a zone?

gigerdo: That's a very good question. I have to double check that since it looks like it wouldn't cover that case.

gigerdo: I think the logic should be (from looking at how the UI does it): 
- Use `node_count_per_zone` if available
- Otherwise estimate it with `size / max_size`

gigerdo: Actually `node_count_per_zone` (Which is supposed to be combined with `memory_per_node`) is unsupported in the provider: It will just ignore those values if set and use the template instead.

So my solution will be to calculate the max from the IC and use that to calculate the nodes-per-zone

gigerdo:
@tobio thank you for the feedback. I implemented your feedback in the last two commits.

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,213 @@
...
+
+		var topology es.ElasticsearchTopologyTF
+		diags := rawTopology.As(ctx, &topology, objectAsOptions)
+		if diags.HasError() {
+			tflog.Debug(ctx, "countNodesInCluster: Failed to read topology.", map[string]interface{}{"error": diags.Errors()})
+			continue
+		}
+
+		size, err := deploymentsize.ParseGb(topology.Size.ValueString())
```
dimuon, line 147: `size` can be omitted in config (so the planned value is "unknown" and parsing fails). I guess we need to take `size` from template in that case.

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,213 @@
...
+		if size > 0 && zoneCount > 0 {
+			nodesInDeployment += zoneCount * nodesPerZone
+		}
+	}
+
+	return nodesInDeployment
+}
+
+func getInstanceConfiguration(template models.DeploymentTemplateInfoV2, topologyId string) *models.InstanceConfigurationInfo {
+	// Find master tier
```
dimuon, line 174: I guess it's not about finding master tier, isn't it?

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
```diff
@@ -0,0 +1,213 @@
...
+	if template.DeploymentTemplate == nil ||
+		template.DeploymentTemplate.Resources == nil ||
+		len(template.DeploymentTemplate.Resources.Elasticsearch) == 0 ||
+		template.DeploymentTemplate.Resources.Elasticsearch[0].Plan == nil {
+		return nil
+	}
+	var masterTier *models.ElasticsearchClusterTopologyElement
```
dimuon, line 181: The same consideration - it's not about master tier.

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
```diff
@@ -483,12 +484,17 @@ func elasticsearchTopologySchema(options topologySchemaOptions) schema.Attribute
...
+	}
+
 	return schema.SingleNestedAttribute{
-		Optional: !options.required,
-		// it should be Computed but Computed triggers TF weird behaviour that leads to unempty plan for zero change config
-		// Computed:    true,
-		Required:    options.required,
-		Description: fmt.Sprintf("'%s' topology element", options.tierName),
+		Optional:      !options.required,
+		Computed:      options.tierName == "master",
```
dimuon, line 494: Does it make sense to make all tiers besides "hot" `Computed`? Otherwise, users have to explicitly specify all tiers which can be scaled up if autoscaling is enabled. I guess, the original TF issue (which was mentioned in the comment) doesn't prevent us from doing that anymore.
If we decide to do this, we need to make sure that `Required` and `Computed` aren't both true.

gigerdo: > the original TF issue (which was mentioned in the comment) doesn't prevent us from doing that anymore.

Not sure if this is the same issue as the comment mentioned, but I did have constant plan changes in node-roles due to a bug in the TF SDK, but I put in a workaround in `node_roles_plan_modifier.go` so it shouldn't happen anymore.

However I don't want to just enable it for all tiers as it isn't necessary for this feature and it may introduce new bugs in certain situations. 

But I think it could be a good change to make in a separate PR. While working on this I noticed also that it could be nice to have the template values already filled in a plan-modifier as well. Then one would see also any tiers and sizes filled in from the template in the plan before applying.

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
```diff
@@ -483,12 +484,17 @@ func elasticsearchTopologySchema(options topologySchemaOptions) schema.Attribute
 		nodeRolesPlanModifiers = append(nodeRolesPlanModifiers, SetUnknownOnTopologySizeChange())
 	}
 
+	var topologyPlanModifiers []planmodifier.Object
+	if options.tierName == "master" {
+		topologyPlanModifiers = append(topologyPlanModifiers, objectplanmodifier.UseStateForUnknown())
```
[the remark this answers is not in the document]
gigerdo: But it makes sense to test the migration change to be sure it works.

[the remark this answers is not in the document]
dimuon: > Hmm probably it doesn't make a difference 🤔 Technically this plan-modifier is not needed, as the other dedicated-master modifier always sets the tier either to a value or null.

I think the new dedicated master modifier can change plan only if `master` tier is not explicitly configured.

ec/ecresource/deploymentresource/plan_modifier.go
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+
+	var plan deploymentv2.DeploymentTF
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	template, err := deptemplateapi.Get(deptemplateapi.GetParams{
```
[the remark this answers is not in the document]
dimuon: Also, does it make sense to check whether `master` tier isn't explicitly configured before making any API calls?

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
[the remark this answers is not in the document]
gigerdo: Yeah but then the field is not unknown anymore and neither of these modifiers applies.

ec/ecresource/deploymentresource/plan_modifier.go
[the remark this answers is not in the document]
gigerdo: Yeah makes sense.

Regarding the template migration: I have to test it, as it isn't clear how it behaves. I think it could lead to problems if the default size for the master tier changes as that update probably wouldn't be applied.

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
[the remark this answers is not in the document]
dimuon: Yeah, indeed.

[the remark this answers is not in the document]
gigerdo: There are probably some corner-cases where the dedicated-master-modifier fails and doesn't apply anything. In those cases it would fall back to `UseStateForUnknown`.

ec/ecresource/deploymentresource/planmodifiers/topology_dedicated_masters.go
gigerdo: I now load size and zone_count from the template as a fallback.

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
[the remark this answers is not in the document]
gigerdo: Actually in my testing I found that I was wrong: The `UseTopologyStateForUnknown` is needed on the master tier as well. Otherwise it will not properly keep the child fields `Unknown`. The reason is that every modifier does actually run, so `UseStateForUnknown` runs first, puts state everywhere. Then the children are not unknown anymore and their plan modifier does not run.

So you were totally right that we need `UseTopologyStateForUnknown` in this case as well.

ec/ecresource/deploymentresource/plan_modifier.go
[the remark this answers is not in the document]
gigerdo: Tested migrating the template, and that works already since the modifier loads the new template and uses the values from there. 
I think the only case that could fail is where the template stays the same, but only the IC changes. Still have to test that case.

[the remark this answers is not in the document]
dimuon: Do we have to test the case when `migrate_to_latest_hardware` is specified? Does the plan contains the new DT in this case?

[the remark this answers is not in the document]
gigerdo: Yeah that I still have to test. I think `migrate_to_latest_hardware` is supposed to keep the same DT, but change to the latest ICs.

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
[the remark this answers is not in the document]
dimuon: Honestly, I forgot about the fact that plan modifiers execute in a certain order which matters.

ec/ecresource/deploymentresource/plan_modifier.go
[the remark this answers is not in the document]
gigerdo: `migrate_to_latest_hardware` works like this:
- By default, every plan just re-uses the values from state (template if there is no value yet, for create)
- `migrate_to_latest_hardware` will reset the value and therefore get the latest value from the template again
- We don't need to look at the `migrateTemplateRequest` as it just contains the full request to update the deployment. But for the dedicated master we anyway only look at the template, which is already loaded separately.

So basically there is currently one problem:
- The modifier will update to the latest IC version even when `migrate_to_latest_hardware` is not set.
- So the modifier should not recalculate the size/zone_count on every plan
- It should only update in the following cases:
  - Node-count has changed and tier should be enabled -> In this case it is fine to use the latest IC
  - `migrate_to_latest_hardware` is set -> Also use the default from the latest IC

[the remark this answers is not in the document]
dimuon: > Node-count has changed and tier should be enabled -> In this case it is fine to use the latest IC
Why? (if `migrate_to_latest_hardware` isn't set?)

So now IIUC, the same template can return the new IC version with other values for `DiscreteSizes` and `MaxZones`. But IMO the dedicated `master` tier plan modifier should use the old values for these fields if `migrate_to_latest_hardware` isn't set.

If this is correct, does it make sense to store`DiscreteSizes` and `MaxZones` in state how we do currently for `InstanceConfigurationId`?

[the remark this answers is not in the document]
gigerdo: Though I tested how it works in the UI, and the UI does always keep the same version. So maybe it is better to not go to the latest IC when enabling the master.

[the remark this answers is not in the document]
dimuon: IIUC the decision to switch to the latest IC should be taken by the user (because it can incur additional costs).
I guess this is one of the reasons behind introducing `migrate_to_latest_hardware` attribute.

> I was originally thinking about using state, but it has some drawbacks: It will also show up in the plan (though the private state could be used for that). And makes everything more complicated because the state is only available in the update-case. So all logic needs to try state, if not available fall back to template.

We can make it read-only and use state for unknown or similar plan modifier (similar to `latest_instance_configuration_id`). I think this is a pretty TF native approach though loading the requested IC version can be a valid alternative.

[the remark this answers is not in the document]
gigerdo: I like the idea of just loading the right IC when needed more. Seems simpler to do one request when needed.

Adding a new field to the state also means having to keep it up-to-date in the read-step for all tiers. So it would still mean having to read it, though it could be skipped if the value is already in the state. And for this specific field I'm not sure if it is very useful for users to see as it is not really the default for the tier (that is set in the tier of the deployment template instead).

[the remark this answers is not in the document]
gigerdo: Though I just realized this IC API only works in ECE 🤔 As it turns out the right way to get them is as part of the deployments API, with the parameter `show_instance_configurations=true`.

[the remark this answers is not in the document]
gigerdo: Another SDK change is needed to be able to get the ICs when loading the deployment: https://github.com/elastic/cloud-sdk-go/pull/469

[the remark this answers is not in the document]
gigerdo: The code currently doesn't compile as it needs another SDK update, but otherwise the PR should now cover everything.

I am now loading the ICs as part of the deployment-read (and storing them in private state), falling back to the template if that isn't available (for create for example).

ec/ecresource/deploymentresource/topology_dedicated_masters.go
```diff
@@ -75,13 +77,21 @@ func UpdateDedicatedMasterTier(
...
-		)
+		if planElasticsearch.MasterTier.IsUnknown() || planElasticsearch.MasterTier.IsNull() {
+			resp.Plan.SetAttribute(ctx,
+				path.Root("elasticsearch").AtName("master"),
+				types.ObjectNull(es.ElasticsearchTopologyAttrs()),
+			)
+		} else {
+			resp.Plan.SetAttribute(ctx,
+				path.Root("elasticsearch").AtName("master").AtName("size"),
+				"0g",
```
dimuon, line 88: Do I get it correctly that  even if `master` tier is defined in plan (because it's, e.g. defined in config), we remove it by setting size to 0? IMO this is the correct behavior.

gigerdo: It is changed if it is already in the plan, not if it is already set in the config (There is a check earlier that stops the logic if the config defines the master).

This happens if the master tier had existed before, for example if it was automatically added. Or if it was configured, but the config was removed. In that case we keep whatever config the plan has and only change the size. 

If I just completely remove the whole tier (by setting it to null) I was running into some problems because the update logic couldn't handle a null master tier if a master had existed previously in the deployment.  But I think this is anyway better as it causes a smaller change and we can keep info about currently used instance-configs in the plan.

ec/ecresource/deploymentresource/topology_dedicated_masters.go
```diff
@@ -0,0 +1,365 @@
...
+			)
+		}
```
dimuon, line 90: nit-pick: it's a quite long if-else, so it makes sense to return from `if` block and remove `else`. I believe that will be more idiomatic golang.

gigerdo: You are right. Another proposal: What if I move the two cases into functions: disable master / enable master. That will make it easier to read

ec/ecresource/deploymentresource/topology_dedicated_masters.go
```diff
@@ -0,0 +1,365 @@
...
+
+		// Skip update if the master tier is already enabled
+		// If migrateToLatestHw is true, update the tier to values from latest IC
+		if masterTierIsEnabled(ctx, planElasticsearch, *template) && !migrateToLatestHw {
+			return
+		}
+
+		// Enable master tier
+
+		instanceConfigurations, diags := ReadPrivateStateInstanceConfigurations(ctx, privateState)
```
dimuon, line 103: Can we use IC from private state if migration to the latest hardware is available or IC or IC version has changed in plan?

ec/ecresource/deploymentresource/topology_dedicated_masters.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	icId := topology.InstanceConfigurationId.ValueStringPointer()
+	if icId == nil || *icId == "" {
+		return nil
+	}
+
+	var deploymentIc *models.InstanceConfigurationInfo
+	for _, ic := range deploymentInstanceConfigs {
+		if ic.ID == *icId {
```
dimuon, line 274: Shall we take `ConfigVersion` into consideration?

ec/ecresource/deploymentresource/plan_modifier.go
```diff
@@ -0,0 +1,56 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	loadTemplate := func() (*models.DeploymentTemplateInfoV2, error) {
+		return deptemplateapi.Get(deptemplateapi.GetParams{
+			API:                        r.client,
+			TemplateID:                 plan.DeploymentTemplateId.ValueString(),
+			Region:                     plan.Region.ValueString(),
+			HideInstanceConfigurations: false,
```
dimuon, line 47: Why not to take ICs from this call instead of reading private state?

ec/ecresource/deploymentresource/topology_dedicated_masters.go
dimuon: Maybe we can if IC version changed (if the IC list in deployment contains all the versions of the used IC) but still does it make sense to read ICs from `loadTemplate` instead (e.g. if template changed)? Also, what shall we do if IC changed?

gigerdo: The private state IC is already not used when migrating, but it isn't straightforward to see in the code (Maybe I can make it more readable):

- The current IC can only be read and used if the tier has an ic-id set
- When migrate-hw is set, the tier is actually set to unknown so we don't know the ic-id (This is due to the `UseTopologyStateForUnknown` modifier)
- since we don't know the ic-id, the code falls back to the template

So it works because `getInstanceConfiguration` below always returns `nil` if `migrateToLatestHw` is set.

gigerdo: Actually this could be made more straightforward by directly using the template if `migrateToLatestHw==true`

ec/ecresource/deploymentresource/plan_modifier.go
gigerdo: The template only contains the last version for each IC. Whereas the deployment-get call contains the IC-versions currently in use (Which are read during 'read' and stored in private state). So we need both sets of ICs: The ones from the private state for regular updates, the template ones for creates and updates with migrate-hw.

ec/ecresource/deploymentresource/topology_dedicated_masters.go
gigerdo: Yeah it is safer. Currently the list of ICs will always contain just one version, but if that part of the API ever changes we still get the right one.

ec/ecresource/deploymentresource/topology_dedicated_masters.go
gigerdo: I refactored the instance-config selection into a separate function and added an if-case for migration.

ec/ecresource/deploymentresource/topology_dedicated_masters.go
gigerdo: Added with the latest commit

ec/ecresource/deploymentresource/topology_dedicated_masters.go
dimuon: > When migrate-hw is set, the tier is actually set to unknown so we don't know the ic-id (This is due to the UseTopologyStateForUnknown modifier)

`migrate-hw` just reflects user's desire to move, the actual possibility to move is defined by this flag and the availability of a newer IC version. IIUC `migrate-hw` can be kept in TF config for some time.

From other side, we can document this behavior. Or check that the actual migration is going to take place.

gigerdo: >  IIUC migrate-hw can be kept in TF config for some time.

Yeah that works fine with the code. The flag basically means 'always use latest IC from template'.
But it doesn't cause any problems in the code. As long as the template has the same IC there will be no changes, as soon as a new IC version is released the next plan will contain changes.

dimuon: @gigerdo , thanks a lot for the explanations.

dimuon (approved):
Great work 👍

gigerdo:
@dimuon thank you very much for taking the time to review this 🥇

gigerdo:
@hnajib-sym we do have plans to release a version 0.11 soon, though we don't have an exact timeline.

nathan-maves:
We also have run into this same issue and would love to see a release soon!

gigerdo:
We have now release 0.11.0 which contains this feature:
https://github.com/elastic/terraform-provider-ec/releases/tag/v0.11.0

PR #817, opened 2024-06-03
Add serverless project resources

tobio:
## Description

~Currently only includes elasticsearch_project resources. IMO we should include at least all project types, also~ likely worthwhile including a serverless_region data source, and maybe project data sources since they should be simple to add re-using the resource read code. 

This PR also include the terraform code generators, currently configured to generate the serverless resource schema and openapi client. 

## How Has This Been Tested?

Manually, acceptance tests. I need to add some more unit tests here. 

## Types of Changes

- [ ] Documentation

## Readiness Checklist

I need to add more tests and docs here.


dimuon:
@tobio , `make gen` adds the license header to a bunch of files.

docs/resources/observability_project.md
```diff
@@ -0,0 +1,87 @@
...
+### Required
+
+- `name` (String) Descriptive name for a project.
+- `region_id` (String) Unique human-readable identifier for a region in Elastic Cloud.
+
+### Optional
+
+- `alias` (String) A custom domain label compatible with RFC-1035 standards. Derived from the project name by default.
+- `optimized_for` (String) The purpose for which the hardware of this elasticsearch project is optimized for. Also known as the Elasticsearch project subtype.
+- `search_lake` (Attributes) Configuration for entire set of capabilities that make the data searchable in Elasticsearch. (see [below for nested schema](#nestedatt--search_lake))
```
dimuon, line 32: `make docs-generate` removes both `optimized_for` and `search_lake` from here as well as from `security_project.md`. It also makes some other changes.

ec/internal/gen/serverless/spec.json
```diff
@@ -0,0 +1,688 @@
+{
```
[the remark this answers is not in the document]
dimuon: Ops, my bad, I mean where to look for the updated `serverless-project-api-dereferenced.yml`.

ec/ecresource/projectresource/mocks.gen.go
```diff
@@ -0,0 +1,259 @@
+// Code generated by MockGen. DO NOT EDIT.
```
dimuon, line 1: How do we generate this file? Obviously I'm missing something, but I didn't manage to regenerate it.

ec/ecresource/projectresource/create_test.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func TestCreate(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	type testData struct {
+		api           api[resource_elasticsearch_project.ElasticsearchProjectModel]
+		modelHandler  modelHandler[resource_elasticsearch_project.ElasticsearchProjectModel]
+		req           resource.CreateRequest
```
dimuon, line 23: IMO `req` can be removed (its value isn't used) with the EXPECT call changed to something like `handler.EXPECT().ReadFrom(ctx, tfsdk.Plan{}).Return(nil, readDiags)` (the actual return values depend on the test case)

ec/ecresource/projectresource/create_test.go
```diff
@@ -0,0 +1,302 @@
...
+				}
+
+				api := NewMockapi[resource_elasticsearch_project.ElasticsearchProjectModel](ctrl)
+				api.EXPECT().Ready().Return(true)
+
+				readDiags := diag.Diagnostics{
+					diag.NewErrorDiagnostic("nope", "nope"),
+				}
+				handler := NewMockmodelHandler[resource_elasticsearch_project.ElasticsearchProjectModel](ctrl)
+				handler.EXPECT().ReadFrom(ctx, req.Plan).Return(nil, readDiags)
```
dimuon, line 47: If `req` is removed, it can become `handler.EXPECT().ReadFrom(ctx, tfsdk.Plan{}).Return(nil, readDiags)`

ec/ecresource/projectresource/create_test.go
```diff
@@ -0,0 +1,302 @@
...
+				},
+			}
+
+			r := Resource[resource_elasticsearch_project.ElasticsearchProjectModel]{
+				api:          td.api,
+				modelHandler: td.modelHandler,
+				name:         "elasticsearch",
+			}
+
+			r.Create(ctx, td.req, &res)
```
dimuon, line 294: If `req` is removed, the line can be changed to `r.Create(ctx, resource.CreateRequest{}, &res)`

ec/ecresource/projectresource/delete_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func TestDelete(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	t.Run("should fail if reading the tf model errors", func(t *testing.T) {
+		ctx := context.Background()
+		req := resource.DeleteRequest{
```
dimuon, line 21: Technically, here and in the second test it can be replaced with `req := resource.DeleteRequest{}`

ec/ecresource/projectresource/elasticsearch_test.go
```diff
@@ -0,0 +1,1163 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+			td := tt.testData(ctx)
+			api := elasticsearchApi{sleeper: fakeSleeper{}}.WithClient(td.client)
+
+			diags := api.Delete(ctx, td.model)
+			require.Equal(t, td.expectedDiags, diags)
+		})
+	}
+}
```
dimuon, line 1163: nice tests 👍

ec/ecresource/projectresource/elasticsearch.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+}
+
+type elasticsearchModelReader struct{}
+
+func (es elasticsearchModelReader) Schema(ctx context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = resource_elasticsearch_project.ElasticsearchProjectResourceSchema(ctx)
+}
+
+func (es elasticsearchModelReader) ReadFrom(ctx context.Context, getter modelGetter) (*resource_elasticsearch_project.ElasticsearchProjectModel, diag.Diagnostics) {
```
dimuon, line 52: IMO we can replace it with a generic, e.g. 
```
func readFrom[T any](ctx context.Context, getter modelGetter) (*T, diag.Diagnostics) {
	var model *T
	diags := getter.Get(ctx, &model)
	return model, diags
}

func (es elasticsearchModelReader) ReadFrom(ctx context.Context, getter modelGetter) (*resource_elasticsearch_project.ElasticsearchProjectModel, diag.Diagnostics) {
  return readFrom[resource_elasticsearch_project.ElasticsearchProjectModel](ctx, getter)
}
```

ec/ecresource/projectresource/elasticsearch.go
```diff
@@ -0,0 +1,364 @@
...
+				fmt.Sprintf("The API request failed with: %d %s\n%s",
+					resp.StatusCode(),
+					resp.Status(),
+					resp.Body),
+			),
+		}
+	}
+
+	model.Id = basetypes.NewStringValue(id)
+	model.Alias = basetypes.NewStringValue(reformatAlias(resp.JSON200.Alias, id))
```
dimuon, line 275: I guess `reformatAlias` doesn't affect `alias` in `Update`, does it?

ec/ecresource/projectresource/read_test.go
```diff
@@ -0,0 +1,187 @@
...
+		expectedId          *string
+	}
+	tests := []struct {
+		name     string
+		testData func(context.Context) testData
+	}{
+		{
+			name: "should fail if reading the tf model errors",
+			testData: func(ctx context.Context) testData {
+				req := resource.ReadRequest{
```
dimuon, line 36: The similar consideration as for create - `req` isn't needed IMO.

dimuon:
@tobio , I completed the review (though I skim through `security_test.go` and `observability_test.go` which look similar to `elasticsearch_test.go`)

ec/ecresource/projectresource/resource.go
```diff
@@ -0,0 +1,131 @@
...
+	api          api[T]
+	name         string
+}
+
+type modelGetter interface {
+	Get(ctx context.Context, target interface{}) diag.Diagnostics
+}
+
+// mockgen doesn't support the recursive generic used within api.WithClient
+// //go:generate go run go.uber.org/mock/mockgen -source=resource.go -destination mocks.gen.go -package projectresource .
```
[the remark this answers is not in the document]
tobio: I've left this here partly to document how the mocks were generated, but also as a reminder to check if the mockgen bug has been fixed. I'll add a link to https://github.com/uber-go/mock/issues/175 as well. 

WDYT?

ec/ecresource/projectresource/mocks.gen.go
tobio: It's [this](https://github.com/elastic/terraform-provider-ec/pull/817#discussion_r1679489861) commented out `go generate` comment

ec/ecresource/projectresource/create_test.go
tobio: I see what you mean. As the code is, the test makes an assertion that "Create calls ReadFrom with the Plan from the `req` parameter". For example, this test would fail if Create altered the Plan in some way prior to calling ReadFrom. 

I'm happy to make the change you're suggesting, this form of assertion has the tendency to be brittle, but the specific value is currently being used, and is providing some additional value to the test. WDYT?

ec/ecresource/projectresource/delete_test.go
tobio: Same as the create test. Using a non-zero value here does provide additional checks during test execution. I'm happy to remove that if we don't think it's worthwhile though.

ec/ecresource/projectresource/elasticsearch.go
tobio: :) I was on the fence about whether this was worthwhile. You + part of me beats the other part of me though.

ec/ecresource/projectresource/elasticsearch.go
tobio: tl;dr No, both Create and Update *must* use the user supplied `alias` property. 

There was some internal discussion about how `alias` is handled by the API, but the end result was that we'll keep the current behaviour for now. 

The API will append `-{short project id}` to whatever alias is specified in the API call, but will return the fully formed alias. So we must consistently send through the user supplied alias without the project id suffix, which means stripping it from the alias returned from the API.

ec/ecresource/projectresource/read_test.go
tobio: Similarly to the Create test. Let's make a decision there and I'll make sure it's consistent across all the tests :)

ec/ecresource/projectresource/resource.go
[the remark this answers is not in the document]
dimuon: Sounds good.

ec/ecresource/projectresource/create_test.go
dimuon: Let's keep it as is.

dimuon (approved):
👍 LGTM

Issue #734, opened 2023-11-15
Setting ec_deployment version via ec_stack data resource with unsupported/legacy topology causes error in the provider


SignalRichard:
## Readiness Checklist
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
When using an Elasticsearch version >= 8.x referencing the `ec_stack` data resource with the legacy `node_type` topology attributes I expect the provider to return the following error: 

`The resources stack version [8.11.1] does not support node_types. Either convert your deployment resource to use node_roles or use a stack version less than [8.0.0]`

## Current Behavior
The following error is instead returned:

```
Error: Value Conversion Error
│
│   with ec_deployment.example_minimal,
│   on main.tf line 39, in resource "ec_deployment" "example_minimal":
│   39:   version                = data.ec_stack.latest.version
│
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│
│ Path: version
│ Target Type: string
│ Suggested Type: basetypes.StringValue
```

## Terraform definition
```
Terraform v1.6.3
on windows_amd64
+ provider registry.terraform.io/elastic/ec v0.9.0
```

## Steps to Reproduce
Run a terraform plan against the following code that shows the expected error and internal error against the 2 `ec_deployment` resources:

```
terraform {
  required_version = "= 1.6.3"

  required_providers {
    ec = {
      source = "elastic/ec"
      version = "0.9.0"
    }
  }

  backend "local" {
    # This should be changed to any other backend configuration, see: https://developer.hashicorp.com/terraform/language/settings/backends/configuration
  }
}

provider "ec" {
  # Configuration options
}


locals {
  ec_region = "azure-eastus"
  stack_version = "8.11.1"
}

# Retrieve the latest stack pack version
data "ec_stack" "latest" {
  version_regex = local.stack_version
  region        = local.ec_region
}

# Create an Elastic Cloud deployment
resource "ec_deployment" "expected_error" {
  # Optional name.
  name = "expected_error"

  region                 = local.ec_region
  version                = local.stack_version 
  deployment_template_id = "azure-general-purpose"

  elasticsearch = {
    hot = {
      node_type_data = "true"
[...]

PR #820, opened 2024-06-18
Bugfix: node-types-validation should handle empty stack version.

gigerdo:
Fixes an error in the validation when using legacy node-type configs.

If the version is unset, the validation crashes with an error. This can happen in certain scenarios: When using a reference (e.g. `data.ec_stack.latest.version`) it will not be resolved in a first phase when this validation is called. In such cases the validation should be skipped.

The validation still works as it is called again after the reading phase when the field has been resolved.

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/734

## Types of Changes

- [ ] Documentation

## Readiness Checklist


dimuon (approved):
LGTM 👍

Issue #802, opened 2024-03-26
Setting empty string for alias parameter generates an error message


mpakoupete:
## Readiness Checklist

- [ ] I am running the latest version
- [ ] I checked the documentation and found no answer
- [ ] I checked to make sure that this issue has not already been filed
- [ ] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

Setting `alias=""` blows up the provider. The provider should likely modify that to `Unknown` during planning.

## Current Behavior

Setting `alias=""` or `alias= null` generates an error message like bellow

```error
(...)
c_deployment.example_minimal: Still creating... [2m40s elapsed]
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to ec_deployment.example_minimal, provider "provider[\"registry.terraform.io/elastic/ec\"]" produced an unexpected new value: .alias:
│ was cty.StringVal(""), but now cty.StringVal("my-example-deployment").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

## Steps to Reproduce

With any [sample](https://registry.terraform.io/providers/elastic/ec/latest/docs/resources/deployment#basic) terraform template by  adding an empty alias `alias=""` we get the error above.

gigerdo:
There is also another problem that even if `alias=""` can be set, the setting is ignored. The `cloud-sdk-go` currently considers `alias=undefined` and `alias=""` to be the same, while the API actually has a different understanding:
- `alias=undefined` -> use default
- `alias=""` -> disable alias

gigerdo:
Will be fixed with release 0.11.0

PR #821, opened 2024-06-19
Bugfix: Correctly handle alias being set to empty value.

gigerdo:
Fixes an error scenario when using the empty string for alias:
- The expectation is to remove the alias
- The empty string used to not be transferred to the backend, so no update to the alias was made
- This lead to a terraform error as the resulting state was different from the config

Now the behavior is:
- Unconfigured -> alias is left untouched and not managed by terraform (When creating this will add a default alias)
- Set to empty string -> alias is removed
- Set to non-empty string -> alias is set to this value

Related change in cloud-sdk-go: https://github.com/elastic/cloud-sdk-go/pull/472

~(Before merging this, we also need a new release of `cloud-sdk-go`. But before creating the release I want to also get a review of the code in case there are additional changes to `cloud-sdk-go` necessary.)~

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/802

## How Has This Been Tested?

Manual testing by adding/removing aliases.

## Types of Changes

- [ ] Documentation

## Readiness Checklist


ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
```diff
@@ -148,9 +148,20 @@ func (plan DeploymentTF) updateBasePayloadsForMigration(client *api.API, newDtId
 }
 
 func (plan DeploymentTF) UpdateRequest(ctx context.Context, client *api.API, state DeploymentTF, migrateTemplateRequest *deployments.MigrateDeploymentTemplateOK) (*models.DeploymentUpdateRequest, diag.Diagnostics) {
+	// The alias behaves like this:
+	// - Not set in the config (null/unknown) -> Alias is not managed by terraform and left untouched
+	// - Set to empty string -> Alias will be removed
+	// - Set to non-empty string -> Alias will be set to this value
+	var alias *string
+	if plan.Alias.IsNull() || plan.Alias.IsUnknown() {
```
tobio, line 156: I think a null alias should be treated the same as `alias = ""` (that is the alias is removed). At this point the user has most likely explicitly set the null value in config.

gigerdo: Yeah I think that would work. My main concern would be that it starts removing the alias for clusters that have an alias but don't have the field defined.

However I think it won't happen due to the field being computed and the `UseStateForUnknown` modifier. So it will always copy from state, which will always either contain the alias or the empty string.

But I'll test it to see if it really behaves properly.

gigerdo: Seems like it won't work: Terraform considers `null` the same as not configuring the field at all (Also documented here: https://developer.hashicorp.com/terraform/language/expressions/types#null)

So even when explicitly setting the config `alias = null` it will end up as an unknown value and the `useStateForUnknownModifier` will set it to the state. 

We could build our own modifier with a different behavior, but since it is documented that `null == not configured` it would then deviate from how terraform usually works.

gigerdo:
The PR now contains the right cloud-sdk-go version

dimuon (approved):
LGTM

Issue #629, opened 2023-04-27
Feature Request: Add RBAC support / user management feature 


pakio:
## Overview

This is the feature request for supporting RBAC and user management feature on this provider.

## Context

Recently Elastic announced that [they started providing the RBAC feature on Elasticsearch service](https://www.elastic.co/guide/en/cloud/current/ec-release-notes-2023-04-18.html).
Though this provider is not yet supporting managing the users by code, this is the feature that many users were wanted for a long time and I suppose there are definitely high demand to manage them in a safer way.

dimuon:
@Kushmaro , I guess this requires adding a dedicated resource to the provider.

Kushmaro:
Yes @dimuon

gigerdo:
This functionality will be included in the upcoming 0.12.0 release.

PR #855, opened 2024-09-09
Implement organization management (RBAC)

gigerdo:
This resource allows managing the members of an organization:

- For each member their roles can be configured.
- Added members will be invited to the organization. 
  - The user will still have to accept the invitation before they have access to the organization.
- Removed members will be removed from the organization

(~Currently all builds/tests for this PR are failing because we first need a new release of cloud-sdk-go to get the organization apis https://github.com/elastic/cloud-sdk-go/pull/477~)

Now includes cloud-sdk-go 1.21.0

## Description

- Regarding invitations
  - The resource tracks the invitation status for each member. Since some time can pass before the new member accepts the invitation. 
  - Since invitations can't be updated, any role changes during this time will mean the invitation is cancelled and a new invitation is created.
  - After the invitation is accepted, the next apply will update the local state with the user-id. And now roles can be updated.
- Create/delete don't really work for an organization. It already exists and can't be deleted (One can only leave an organization).
  - The default flow is to import the already existing organization using the `terraform import` flow.
  - It isn't possible to automatically import with the create, because then the resulting organization potentially wouldn't match the current config. So the create has to throw an error if it is called.
  - The delete does nothing currently, so it is possible to run the `terraform destroy` to remove the org from the state. However this is currently just to stop managing the resource in terraform, it won't actually be deleted.

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/629

## How Has This Been Tested?

- Manual testing
- There are no real acceptance tests because it isn't possible to automatically add new members.
- However currently all use-cases are covered by acceptance tests that use mocked API requests.

## Types of Changes

- [x] Documentation

## Readiness Checklist


tobio (commented):
I wonder if we'd have a better UX restructuring this one a bit:

* An organisation data source. Like you've said there's no real configuration for an organisation, moving this to a data source would avoid the import requirement
* Having a resource per organisation member, rather than a single resource managing all members within an organisation. 

The TF module could look something like:

```hcl
data "ec_organization" "myorg" {}

resource "ec_organization_member" "tobio" {
  organization_id = data.ec_organization.myorg.id
  email = "tobio@example.com"
  deployment_roles = [...]
  ...
}
```

Managing multiple members with the same permissions becomes simpler:

```hcl
resource "ec_organization_member" "admins" {
  for_each = ["tobio@example.com", "gigerdo@example.com"]
  organization_id = ...
  email = each.key
  deployment_roles = [...]
```

gigerdo:
I think both have their advantages. The main difference in my opinion is on the management of members:

- Single organization resource: You must manage all members. This has the advantage that you can always be sure that the configured members are the only ones in the org.
- Per-member resource: More flexible, you can pick which members should be managed by terraform.

gigerdo:
The PR is now updated to include the right cloud-sdk-go and should compile now.

gigerdo:
Small update to the PR, I reenamed the fields:
- `for_all_deployments` -> `all_deployments`
-  `for_all_projects` -> `all_projects`

ec/ecresource/organizationresource/mapper_to_model.go
```diff
@@ -0,0 +1,194 @@
...
+func nilToEmpty(id *string) string {
+	if id == nil {
+		return ""
+	}
+	return *id
+}
+
+func organizationRoleApiToModel(member models.OrganizationMembership) types.String {
+	if member.RoleAssignments != nil &&
+		member.RoleAssignments.Organization != nil &&
```
dimuon, line 76: It can be omitted - the next line checks `len` anyway
```suggestion
```

ec/ecresource/organizationresource/mapper_to_model.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/elastic/cloud-sdk-go/pkg/models"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func apiToModel(ctx context.Context, member models.OrganizationMembership, invitationPending bool, diagnostics *diag.Diagnostics) *OrganizationMember {
+	organizationRole := organizationRoleApiToModel(member)
+	if diagnostics.HasError() {
```
dimuon, line 31: Is it a typo? (Do we need to check diagnostics here?)

ec/ecresource/organizationresource/mapper_roles.go
```diff
@@ -0,0 +1,40 @@
...
+import "strings"
+
+type RoleType string
+
+const (
+	Deployment           = "deployment"
+	ProjectElasticsearch = "elasticsearch"
+	ProjectObservability = "observability"
+	ProjectSecurity      = "security"
+)
```
dimuon, line 29: IIUC, we don't need to export these definitions:
```suggestion
	deployment           = "deployment"
	projectElasticsearch = "elasticsearch"
	projectObservability = "observability"
	projectSecurity      = "security"
)
```

ec/ecresource/organizationresource/mapper_to_model.go
```diff
@@ -0,0 +1,194 @@
...
+
+		result = append(result, ProjectRoleAssignment{
+			Role:             types.StringValue(roleApiToModel(*roleAssignment.RoleID, roleType)),
+			ForAllProjects:   forAllApiToModel(roleAssignment.All),
+			ProjectIDs:       projectIds,
+			ApplicationRoles: applicationRoles,
+		})
+	}
+
+	roleAssignments, diags := types.SetValueFrom(ctx, setElementObjectType(schema), result)
```
dimuon, line 180: IMO we can simplify this:
```suggestion
	roleAssignments, diags := types.SetValueFrom(ctx, schema.GetNestedObject().GetAttributes().Type(), result)
```

ec/ecresource/organizationresource/mapper_to_model.go
```diff
@@ -0,0 +1,194 @@
...
+
+			result = append(result, DeploymentRoleAssignment{
+				Role:              types.StringValue(roleApiToModel(*roleAssignment.RoleID, Deployment)),
+				ForAllDeployments: forAllApiToModel(roleAssignment.All),
+				DeploymentIDs:     deploymentIds,
+				ApplicationRoles:  applicationRoles,
+			})
+		}
+	}
+	roleAssignments, diags := types.SetValueFrom(ctx, setElementObjectType(deploymentRoleAssignmentsSchema()), result)
```
dimuon, line 110: nit-pick: maybe change it to this?
```suggestion
	roleAssignments, diags := types.SetValueFrom(ctx, deploymentRoleAssignmentsSchema().NestedObject.GetAttributes().Type(), result)
```

dimuon:
@gigerdo , does it make sense to add unit tests to cover negative use cases? 
(IIUC, `TestOrganizationResourceAgainstMockedAPI` covers positive use cases only).

ec/ecresource/organizationresource/update.go
```diff
@@ -0,0 +1,291 @@
...
+	stateMember OrganizationMember,
+	planMember OrganizationMember,
+	organizationID string,
+	diagnostics *diag.Diagnostics,
+) {
+	if planMember.InvitationPending.ValueBool() {
+		// Invitations can't be updated, so while the invitation is pending the role assignments can't be changed
+		// The only way to update them is by creating a new invitation with the right role-assignments.
+		r.deleteInvitation(planMember, organizationID, diagnostics)
+		r.createInvitation(ctx, stateMember.Email.ValueString(), planMember, organizationID, diagnostics)
```
dimuon, line 114: Just for my understanding: I guess `stateMember.Email` should equal to `planMember.Email`, correct?

ec/ecresource/organizationresource/update.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	remove := models.ProjectRoleAssignments{
+		Elasticsearch: removeElasticsearch,
+		Observability: removeObservability,
+		Security:      removeSecurity,
+	}
+	return &add, &remove
+}
+
+func difference[T interface{}](a, b []*T, getKey func(T) string) []*T {
```
dimuon, line 269: IMO we need unit test for this func (as for other funcs (at least internal ones) in this package).

dimuon: Does it make sense to use something like https://github.com/hashicorp/go-set/blob/main/hashset.go instead?

ec/ecresource/organizationresource/schema.go
```diff
@@ -0,0 +1,250 @@
...
+					Required:            true,
+				},
+				"all_deployments": schema.BoolAttribute{
+					MarkdownDescription: "Role applies to all deployments in the organization.",
+					Optional:            true,
+					PlanModifiers: []planmodifier.Bool{
+						planmodifiers.BoolDefaultValue(false), // consider unknown as false
+					},
+				},
+				"deployment_ids": schema.SetAttribute{
```
dimuon, line 140: What happen if the same deployment ID is used in 2 different nested objects? e.g.
first object:
``` 
        {
          role           = "viewer"
          deployment_ids = ["1111"]
        }
```
second object:
``` 
        {
          role           = "admin"
          deployment_ids = ["1111"]
        }
```
Does it make sense to validate cases like this?

ec/ecresource/organizationresource/mapper_to_model.go
gigerdo: Yeah, I think `organizationRoleApiToModel` originally could throw an error and I forgot to remove the check.

ec/ecresource/organizationresource/update.go
gigerdo: Yeah I will check if that simplifies the code

ec/ecresource/organizationresource/update.go
gigerdo: Yeah, the email is a computed read-only field which will never change. The plan is always the same as the state due to `UseStateForUnknown`. And it is always the same as the key in the map of members.

It would also be possible to always pass through the config-key of the map and use that. Since that is actually the source-of-truth for which member this is.

gigerdo: I think using the same email (the one set by the user in the config) is the way to go. This way we always use the same source for all update calls (Instead of sometimes using the key and sometimes using the email field in the value of the map)

ec/ecresource/organizationresource/resource_test.go
```diff
@@ -0,0 +1,646 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "members.%", "1"),
+					resource.TestCheckNoResourceAttr(resourceName, "members.newuser@example.com"),
+				),
+			},
+		},
+	})
+}
+
+func mockApi() *api.API {
```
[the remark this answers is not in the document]
gigerdo: I think the main mock function is quite important for the test: The list of mocked responses is chosen to match the list of testcases, so every testcase has their exact mocks.

However I could see moving mocking to another file being better for readability, since it makes it easier to have the mocks open in a separate view.

ec/ecresource/organizationresource/mapper_to_model.go
gigerdo: Yeah that's simpler :+1:

ec/ecresource/organizationresource/update.go
dimuon: Yeah, I suppose we can proceed with the difference function, just cover it by tests.

ec/ecresource/organizationresource/resource_test.go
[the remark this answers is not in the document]
gigerdo: I'm now also thinking if I should split the tests into separate cases. Since they are all mocked anyway, there is no need to have one huge testcase.

ec/ecresource/organizationresource/update.go
gigerdo: :+1: I'm adding some tests for the difference function

ec/ecresource/organizationresource/resource_test.go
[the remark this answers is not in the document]
gigerdo: I split up all the cases, this will also make it much easier to investigate what is actually failing (though it does add some boilerplate)

gigerdo:
@dimuon I refactored the testcases into separate tests, and added a failure case for each.

ec/ecresource/organizationresource/resource_mock_utils_test.go
dimuon, line 1: nit-pick: usually files with names like `*_test.go` contain test functions while this one contains test utility funcs. Does it make sense to rename it? (e.g. `resource_mock_test_utils.go`)

dimuon (approved):
LGTM 👍

ec/ecresource/organizationresource/resource_mock_utils_test.go
gigerdo: I also don't quite like this name. But the problem is that if the file doesn't end with _test, it must be part of the regular code (and package). If you end files with _test, it is allowed to put them into a separate test package. So if I want the mock code to be in the test package it unfortunately has to be in a _test file.

Though it isn't necessary to have a separate package for tests, I do like it since it ensures the testing utils can't accidentally be called from regular code.

Issue #854, opened 2024-09-09
Resizing a deployment with terraform reset the default cloud snapshot policy to its defaults


frconil:
## Readiness Checklist

- [ x ] I am running the latest version
- [ x ] I checked the documentation and found no answer
- [ x ] I checked to make sure that this issue has not already been filed
- [ x ] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

Given a manifest where I define both the size of the deployment (e.g. number of nodes for a given tier) and update the default SLM policy, I would expect the SLM policy to stay the same across terraform runs if I do not modify the SLM policy, unless I changed settings that are not meant to be changed.

## Current Behavior

If I modify the number of nodes, the first run also (silently) resets the SLM policy to the cloud defaults. The second `terraform apply` will update the SLM policy as per the terraform definition. 

## Terraform definition

```
terraform {
  required_version = ">= 1.0.0"
  required_providers {
    ec = {
      source  = "elastic/ec"
    }
     elasticstack = {
      source  = "elastic/elasticstack"
   }
  }
}

provider "ec" {
apikey = "REDACTED"
}

resource "ec_deployment" "custom-deployment" {
  name                   = "My deployment identifier"
  region                 = "gcp-europe-west3"
  version                = "8.15.0"
  deployment_template_id = "gcp-memory-optimized-v2"

 elasticsearch = {
    hot = {
      size = "4g"
      zone_count="3"
      autoscaling = {}
    }
  }
  kibana = {}
}

provider "elasticstack" {
  elasticsearch {
    username = ec_deployment.custom-deployment-fc.elasticsearch_username
    password = ec_deployment.custom-deployment-fc.elasticsearch_password
  endpoints = ["${ec_deployment.custom-deployment-fc.elasticsearch.https_endpoint}"]
  }
}

resource "elasticstack_elasticsearch_snapshot_lifecycle" "cloud-snapshot-policy" {
  name = "cloud-snapshot-policy"
  schedule      = "0 0 1 * * ?"
  snapshot_name = "<cloud-snapshot-{now/d}>"
  repository    = "found-snapshots"
  include_global_state = true
  expire_after = "30d"
  min_count    = 5
  max_count    = 50
}
```

## Steps to Reproduce

1. Using the manifest above, update the `zone_count` parameter up or down. 
2. Run terraform apply twice
3. first run will return:

```
Terraform will perform the following actions:

  # ec_deployment.custom-deployment will be updated in-place
[...]

PR #858, opened 2024-09-20
Bugfix: Only pass snapshot settings into the update plan if there are changes

gigerdo:
- This ensures the settings are only put into the update request if there are actual changes in the config
- Otherwise the settings are left out of the payload
- This solves a bug where every apply would reset the policy back to the default (because putting the default settings in the plan will cause it to reset the policy).
  - When using the elasticstack provider to manage the policy, we want the ec provider to not touch it.

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/854
## Types of Changes

- [ ] Documentation

## Readiness Checklist


nick-benoit (approved):
👍 Nice work. Nice tests as well!

nick-benoit (approved):
LGTM

Issue #842, opened 2024-08-20
Terraform provider corrupts trust settings on update


nick-benoit:
## Readiness Checklist

- [x] I am running the latest version
- [ ] I checked the documentation and found no answer
- [ ] I checked to make sure that this issue has not already been filed
- [ ] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

Trust settings should not be altered unless specified in the terraform file. 

## Current Behavior

Elasticsearch trust settings are changed by the provider in the following unexpected ways:

* Account trust names are removed
* Direct trust is removed 

An example of Elasticsearch cluster data before TF update:

```
 "trust": {
    "accounts": [
      {
        "account_id": "554443047",
        "name": "Default trust",
        "trust_all": true
      }
    ],
    "external": [],
    "direct": [
      {
        "uid": "9a2d84bb-0619-4e2c-8da1-1151e34ab890",
        ...
        "type": "generic",
        "name": "Test"
      }
    ]
  },
```

And after TF Update

```
  "trust": {
    "accounts": [
      {
        "trust_all": true,
        "account_id": "554443047",
        "name": "" # This was removed 
      }
    ],
    "external": [],
    "direct": [] # This was also removed
  },
```

## Terraform definition

```
terraform {
  required_version = ">= 0.12.29"

  required_providers {
    ec = {
      source  = "elastic/ec"
      version = "0.10.0"
    }
  }
}

provider "ec" {
  # ECE installation endpoint
  endpoint = "https://public-api.qa.cld.elstc.co"

  # If the ECE installation has a self-signed certificate
  # setting "insecure" to true is required.
  #insecure = false

  # APIKey is the recommended authentication mechanism. When
  # Targeting the Elasticsearch Service, APIKeys are the only
  # valid authentication mechanism.
  apikey = "..."
}

data "ec_stack" "latest" {
  version_regex = "latest"
  region        = "aws-eu-west-1"
}

# Create an Elastic Cloud deployment
resource "ec_deployment" "example_minimal" {
  # Optional name.
  name = "my_example_deployment"

  # Mandatory fields
  region                 = "aws-eu-west-1"
  version                = "8.14.3" #data.ec_stack.latest.version
  deployment_template_id = "aws-io-optimized-v2"

  # Use the deployment template defaults
  elasticsearch = {
    hot = {
      autoscaling = {}
    }

    ml = {
       autoscaling = {
          autoscale = true
       }
    }

  }

  kibana = {
[...]

PR #859, opened 2024-09-20
Elasticsearch trust settings: Only add them to the update payload if there are changes.

gigerdo:
The trust setting should only be added to the payload when they change. If they are left out of the payload, no changes to the trust will be made.

This is a partial fix/workaround for https://github.com/elastic/terraform-provider-ec/issues/842 

## Related Issues

https://github.com/elastic/terraform-provider-ec/issues/842

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_payload.go
+++ b/ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_payload.go
@@ -61,14 +61,6 @@ func ElasticsearchPayload(ctx context.Context, plan types.Object, state *types.O
 		return nil, diags
 	}
 
-	var esState *ElasticsearchTF
-	if state != nil {
-		esState, diags = objectToElasticsearch(ctx, *state)
-		if diags.HasError() {
-			return nil, diags
-		}
-	}
-
 	if es == nil {
 		return nil, nil
 	}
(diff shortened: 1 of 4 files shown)


ec/ecresource/deploymentresource/elasticsearch/v2/elasticsearch_payload.go
```diff
@@ -160,11 +168,14 @@ func (es *ElasticsearchTF) payload(ctx context.Context, res *models.Elasticsearc
 		res.Plan.AutoscalingEnabled = ec.Bool(es.Autoscale.ValueBool())
 	}
 
-	res.Settings, ds = elasticsearchTrustAccountPayload(ctx, es.TrustAccount, res.Settings)
-	diags.Append(ds...)
+	// Only add trust settings to update payload if trust has changed
+	if state == nil || !es.TrustAccount.Equal(state.TrustAccount) || !es.TrustExternal.Equal(state.TrustExternal) {
```
nick-benoit, line 172: Is `state == nil` the bootstrapping case? In either case, does it make sense to write these trust values in that case vs just relying on the deployment defaults?

gigerdo: Yep that's the creation case. But we can't just skip it, it could be that the user has already added trust configs when creating the deployment. If the user has not configured anything, the plan will be empty so it won't add any payloads either (But this is handled inside the payload functions)

dimuon (approved):

nick-benoit (approved):

PR #878, opened 2024-11-21
bugfix: don't use nil state for a computed attribute

gersonsosa:
## Description

If a computed attribute parent state is null, don't try to use the state to avoid inconsistency errors when the apply process finishes the state will still be used in the previous cases to preserve the same functionality.

https://elasticco.atlassian.net/browse/CP-9334

## Motivation and Context

We found a bug while fixing the acceptance tests where multiple stateless components have this inconsistent behavior, in the eyes of terraform this is a bug.

> The New State object returned from the provider must meet the following constraints:
> * Any attribute that had a known value in the Final Planned State must have an identical value in the new state. In particular, if the remote API returned a different serialization of the same value then the provider must preserve the form the user wrote in the configuration, and must not return the normalized form produced by the provider.

See the docs for more details https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md#applyresourcechange

## Types of Changes

- [ ] Documentation

## Readiness Checklist

### DIFF
--- a/ec/internal/planmodifiers/use_state_unless_template_changed.go
+++ b/ec/internal/planmodifiers/use_state_unless_template_changed.go
@@ -65,8 +65,7 @@ func (m useStateForUnknownUnlessMigrationIsRequired) UseState(ctx context.Contex
 	var diags diag.Diagnostics
 
 	var parentResState attr.Value
-	diag := state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState)
-	diags.Append(diag...)
+	diags.Append(state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState)...)
 
 	resourceIsBeingCreated := parentResState.IsNull()
 
--- a/ec/internal/planmodifiers/use_state_unless_template_changed.go
+++ b/ec/internal/planmodifiers/use_state_unless_template_changed.go
@@ -65,7 +65,9 @@ func (m useStateForUnknownUnlessMigrationIsRequired) UseState(ctx context.Contex
 	var diags diag.Diagnostics
 
 	var parentResState attr.Value
-	diags.Append(state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState)...)
+	if d := state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState); d.HasError() {
+		return false, d
+	}
 
 	resourceIsBeingCreated := parentResState.IsNull()
 
--- a/ec/internal/planmodifiers/use_state_unless_template_changed.go
+++ b/ec/internal/planmodifiers/use_state_unless_template_changed.go
@@ -31,6 +31,7 @@ import (
 //  1. The attribute is not nullable (`isNullable = false`) and the topology's state is nil
 //  2. The deployment template attribute has changed
 //  3. `migrate_to_latest_hardware` is set to `true` and there is a migration available to be performed
+//  4. The state of the parent attribute is nil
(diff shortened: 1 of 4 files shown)


ec/internal/planmodifiers/use_state_unless_template_changed.go
```diff
@@ -63,6 +64,16 @@ func (m useStateForUnknownUnlessMigrationIsRequired) PlanModifyInt64(ctx context
 func (m useStateForUnknownUnlessMigrationIsRequired) UseState(ctx context.Context, configValue attr.Value, plan tfsdk.Plan, state tfsdk.State, planValue attr.Value, stateValue attr.Value) (bool, diag.Diagnostics) {
 	var diags diag.Diagnostics
 
+	var parentResState attr.Value
+	diag := state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState)
+	diags.Append(diag...)
```
dimuon, line 69: We need to either check `diag` for an error and return it if it contains one or return it instead of `nil` in all subsequent `return` statements. (the method uses the second way so it accumulates all encountered errors in `diags`).
As a side note, name `diag` hides the corresponding imported package so IMO it's better to rename the error var or better do something like `diags.Append(state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState)...)`

dimuon:
@gersonsosa , the PR looks good besides one comment.

ec/internal/planmodifiers/use_state_unless_template_changed.go
gersonsosa: thank you for your comment, indeed that would cause an issue, in this case it would be unexpected that this would generate an error so is tempting to check and return, but again since this is unlikely to happen I think I'll follow the same pattern I'll add the diagnostics to the `diags` variable

ec/internal/planmodifiers/use_state_unless_template_changed.go
```diff
@@ -63,6 +64,15 @@ func (m useStateForUnknownUnlessMigrationIsRequired) PlanModifyInt64(ctx context
 func (m useStateForUnknownUnlessMigrationIsRequired) UseState(ctx context.Context, configValue attr.Value, plan tfsdk.Plan, state tfsdk.State, planValue attr.Value, stateValue attr.Value) (bool, diag.Diagnostics) {
 	var diags diag.Diagnostics
 
+	var parentResState attr.Value
+	diags.Append(state.GetAttribute(ctx, path.Root(m.resourceKind), &parentResState)...)
+
+	resourceIsBeingCreated := parentResState.IsNull()
+
+	if resourceIsBeingCreated {
+		return false, nil
```
dimuon, line 73:
```suggestion
		return false, diags
```
The same should be applied down the lines in all places where currently `nil` is returned.

dimuon: I mean we should return `diags` also in lines 77, 81 and 86.

dimuon: Sorry if I failed to make myself clear earlier.

gersonsosa: mm should I only add it here? why the others doesn't have it

gersonsosa: maybe is better to check and return, this looks a bit confusing

dimuon (approved):
LGTM

PR #939, opened 2025-11-06
populate admin_features_package and product_types in ec_security_project ReadI()

### LINKED ISSUE #938: [BUG] ec_security_project - provider returns an error indicating that `admin_features_package` and `product_types` are unknown values
## Readiness Checklist
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

After successfully applying an `ec_security_project` resource, all attributes should have known values in the Terraform state. The `admin_features_package` and `product_types` attributes should be populated with their values from the API response.

## Current Behavior

After running `terraform apply` on an `ec_security_project` resource, the provider returns an error indicating that `admin_features_package` and `product_types` are still unknown values:

```
ec_security_project.test: Still creating... [1m20s elapsed]
╷
│ Error: Provider returned invalid result object after apply
│
│ After the apply operation, the provider still indicated an unknown value for ec_security_project.test.admin_features_package. All values must be known after apply, so this is always a bug in the provider and should be
│ reported in the provider's own repository. Terraform will still save the other known object values in the state.
╵
╷
│ Error: Provider returned invalid result object after apply
│
│ After the apply operation, the provider still indicated an unknown value for ec_security_project.test.product_types. All values must be known after apply, so this is always a bug in the provider and should be reported in
│ the provider's own repository. Terraform will still save the other known object values in the state.
╵
```

## Terraform definition

```hcl
resource "ec_security_project" "test" {
  name      = "my-security-project"
  region_id = "aws-us-east-1"
  # potentially includes admin_features_package and/or product_types
}
```

## Steps to Reproduce

1. Create a Terraform configuration with an `ec_security_project` resource
2. Set `admin_features_package` and/or `product_types` attributes (or rely on defaults)
3. Run `terraform apply`
4. Observe the error about unknown values after apply

## Context

This prevents successful creation of security projects via Terraform. The resource is created on the Elastic Cloud side, but Terraform fails to complete the apply operation cleanly, leaving the state in a partially updated condition.

## Possible Solution

Looking at `ec/ecresource/projectresource/security.go`:

The `Read()` function (lines 225-292) successfully reads most attributes from the API response but does not populate the `admin_features_package` or `product_types` fields from `resp.JSON200`. These fields need to be read from the API response and set on the model, similar to how other attributes like `name`, `region_id`, `alias`, etc. are handled.

The fields are sent during `Create()` (lines 110-112 for `admin_features_package` and lines 114-130 for `product_types`), but they're never read back from the API response in the `Read()` function, causing them to remain unknown after the create operation completes.

## Your Environment

* Version used: 0.12.2-dev (current master branch)
* Running against Elastic Cloud SaaS or Elastic Cloud Enterprise and version: Elastic Cloud SaaS
* Environment name and version: Terraform v1.x+, Go 1.21+
* Operating System and version: macOS/Linux

norrietaylor:
The Read() function successfully reads most attributes from the API response but does not populate the admin_features_package or product_types fields from resp.JSON200. These fields need to be read from the API response and set on the model, similar to how other attributes like name, region_id, alias, etc. are handled.

The fields are sent during Create(), but they're never read back from the API response in the Read() function, causing them to remain unknown after the create operation completes.

This patch addresses this by adding by implementing ProductTypesListType and ProductTypesListValue they:
  - Contain ListSemanticEquals() method for order-insensitive comparison
  - Compare product_types as sets (by product_line/product_tier pairs) rather than ordered lists

Appropriate unit and acceptance tests have been added. 

   Fixes #938

   Assisted by Cursor AI.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -25,7 +25,7 @@ jobs:
         terraform_wrapper: false
 
     - name: Cache Go Modules
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: ~/go/pkg/mod
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}


ec/ecresource/projectresource/security.go
```diff
@@ -43,8 +47,77 @@ func NewSecurityProjectResource() *Resource[resource_security_project.SecurityPr
 
 type securityModelReader struct{}
 
+// productTypesOrderInsensitivePlanModifier ignores order differences in product_types list
+type productTypesOrderInsensitivePlanModifier struct{}
```
tobio: Based on the comment here, I think we actually want to create a [custom type](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom) for the product types. That type can then implement [`ListValuableWithSemanticEquals`](https://github.com/hashicorp/terraform-plugin-framework/blob/7fb50058928e09bae9289d436c60b576270d1bf5/types/basetypes/list_value.go#L32) which:
* Applies the same logic as this plan modifier during the planning phase
* Critically, applies the same logic during the post application checks which will avoid the inconsistent state error I imagine we'd get if the TF config defined product types in a different order to what was returned by the API.

tobio (approved):

Issue #954, opened 2025-12-16
ec_organization: Inconsistent state error with multiple deployment IDs in deployment_roles


omer2500:
## Readiness Checklist
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
When applying an `ec_organization` resource with members that have deployment roles containing multiple deployment IDs, the provider should successfully apply the changes without producing inconsistent state errors. The provider should be able to correctly compare the planned state with the actual state returned from the API.

## Current Behavior
The provider fails with a "Provider produced inconsistent result after apply" error when managing organization members with deployment roles that contain multiple deployment IDs or application roles. This occurs because:

1. When sending data to the API (`mapper_to_api.go`), the provider sorts `deployment_ids` and `application_roles` arrays
2. When reading data back from the API (`mapper_to_model.go`), the provider does NOT sort these arrays
3. The API returns the arrays in a different order than they were sorted before sending
4. Terraform's set comparison sees them as different elements, causing the inconsistent result error

**Error message:**
```
Error: Provider produced inconsistent result after apply

When applying changes to ec_organization.example, provider
"provider["registry.terraform.io/elastic/ec"]" produced an unexpected new
value: .members["user@example.com"].deployment_roles: planned set element
cty.ObjectVal(map[string]cty.Value{"all_deployments":cty.False,
"application_roles":cty.NullVal(cty.Set(cty.String)),
"deployment_ids":cty.SetVal([]cty.Value{cty.StringVal("a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6"),
cty.StringVal("b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6q7"),
cty.StringVal("c3d4e5f6g7h8i9j0k1l2m3n4o5p6q7r8"),
cty.StringVal("d4e5f6g7h8i9j0k1l2m3n4o5p6q7r8s9")}),
"role":cty.StringVal("admin")}) does not correlate with any element in
actual.

This is a bug in the provider, which should be reported in the provider's
own issue tracker.
```

This affects anyone managing organization members with:
- Multiple deployment IDs in a deployment role assignment
- Multiple application roles in a deployment or project role assignment
- Multiple project IDs in a project role assignment

## Terraform definition

```hcl
locals {
  deployments = yamldecode(file("../deployments.yaml"))
[...]

PR #957, opened 2025-12-18
Remove old role assignments first

tobio:
Fixes https://github.com/elastic/terraform-provider-ec/issues/954

This PR reverses the order we handle role assignment changes in the organisation resource. 
Currently if the existing and proposed roles overlap (for example existing includes deployment a, whilst proposed includes deployments a and b) then the overlapping deployments will be removed from the role assignments, resulting in the role assignments not being correctly reconciled, and an inconsistent state error.

By reversing the order, we ensure than any overlapping deployments/projects are included in the final permission set. 

Is this the 'best' approach, well no, it would be preferable to explicitly add/remove deployments that are being removed from the permission set. That's a much bigger change to the resource though, and this simplified approach works pretty well. The concern here, is that if Terraform fails between the remove/add then an organisation member will be left without access to some deployments they should have access to, arguably this is safer than the alternative where a member will have access to deployments they should not have access to 🤷

### DIFF
--- a/examples/deployment_ec2_instance/provider.tf
+++ b/examples/deployment_ec2_instance/provider.tf
@@ -9,7 +9,7 @@ terraform {
 
     aws = {
       source  = "hashicorp/aws"
-      version = "0.12.4"
+      version = "~> 5.45, != 5.71.0"
     }
   }
 }


dimuon (approved):

dimuon (approved):
It will be nice to have a test (maybe unit test) to cover the issue.

omer2500:
@tobio Thanks for the quick fix! we installed the provider locally from you fork and it worked as expected
waiting it to get released :)

tobio:
> It will be nice to have a test (maybe unit test) to cover the issue.

Added in https://github.com/elastic/terraform-provider-ec/pull/957/commits/44b8479ac3c9305119034828fd130c5f8961f605

PR #992, opened 2026-04-29
fix: mark rule IDs unknown when ec_deployment_traffic_filter is renamed

### LINKED ISSUE #991: ec_deployment_traffic_filter: "Provider produced inconsistent result after apply" when renaming a filter
## Description

When an `ec_deployment_traffic_filter` resource is renamed (only the `name` attribute changes), Terraform fails on apply with:

```
Error: Provider produced inconsistent result after apply
...planned set element does not correlate with any element in actual
```

All subsequent `terraform apply` runs fail until the workspace is manually repaired.

## Root Cause

When a traffic filter is renamed, the Elastic Cloud API re-assigns **new IDs to all existing rules**. The `StringIsUnknownIfRulesChange` plan modifier (in `schema.go`) correctly handles rule additions/removals by marking rule `id` attributes as unknown — but it does not account for a **name-only change**.

As a result, Terraform plans with the old (stale) rule IDs, the API returns new ones, and the set reconciliation fails.

There is already a comment in the code acknowledging that IDs change on update:
```go
// NOTE: The ID will change on update, so we intentionally do not use plan modifier resource.UseStateForUnknown() here!
```

…but `PlanModifyString` only checks for rule additions/removals — the name-change case was missed.

## Steps to Reproduce

```hcl
resource "ec_deployment_traffic_filter" "this" {
  name   = "my-filter-renamed"   # was "my-filter"
  region = "gcp-us-west2"
  type   = "ip"
  rule { source = "1.2.3.4/32" }
}
```

1. `terraform apply` with `name = "my-filter"` → succeeds
2. Change to `name = "my-filter-renamed"` (rules unchanged)
3. `terraform apply` → fails with "Provider produced inconsistent result after apply"

## Expected Behaviour

The rename succeeds and Terraform correctly reconciles the new rule IDs returned by the API.

## Suggested Fix

In `PlanModifyString`, before the existing rule-change detection, add a name-change check: if `name` differs between state and plan, mark the rule `id` as unknown. This causes Terraform to plan with `(known after apply)` rather than stale IDs, which accepts any concrete value the API returns.

```go
// If the filter name changed, the API re-assigns new IDs to all existing rules.
var stateName, planName types.String
resp.Diagnostics = append(resp.Diagnostics, req.State.GetAttribute(ctx, path.Root("name"), &stateName)...)
resp.Diagnostics = append(resp.Diagnostics, req.Plan.GetAttribute(ctx, path.Root("name"), &planName)...)
if resp.Diagnostics.HasError() {
    return
}
if !stateName.Equal(planName) {
    resp.PlanValue = types.StringUnknown()
    return
}
```

No new imports needed — `path` and `types` are already imported.

## Affected Versions

Confirmed on v0.12.4 and v0.12.5. Non-rename operations (add rule, remove rule, update description) are unaffected.

---

*This issue was identified and investigated with the assistance of [Claude](https://claude.ai) (Anthropic).*

daverick:
## Description

When an `ec_deployment_traffic_filter` resource is renamed (only `name` changes), the Elastic Cloud API re-assigns new IDs to all existing rules. The `StringIsUnknownIfRulesChange` plan modifier already handles rule additions/removals by marking rule `id` as unknown, but missed the name-only change case — causing Terraform to plan with stale rule IDs and fail with "Provider produced inconsistent result after apply".

This fix adds a name-change check before the existing rule-change detection: if `name` differs between state and plan, all rule IDs are marked unknown so Terraform plans with `(known after apply)` rather than stale IDs.

## Related Issues

Fixes #991

## Motivation and Context

Users who rename an `ec_deployment_traffic_filter` resource hit a hard blocker: the apply fails and the workspace is broken until manually repaired. Non-rename operations (add/remove rules, update description) are unaffected.

There is already a comment in the code acknowledging IDs change on update:
```go
// NOTE: The ID will change on update, so we intentionally do not use plan modifier resource.UseStateForUnknown() here!
```
…but `PlanModifyString` only checked for rule additions/removals — the name-change path was missed.

## How Has This Been Tested?

Added `TestResourceTrafficFilter_nameRename` unit test in `resource_test.go`. The test:
1. Creates a filter with two rules (rule IDs `some-random-rule-id-1`, `some-random-rule-id-2`)
2. Renames it — mock returns **new** rule IDs (`some-random-rule-id-3`, `some-random-rule-id-4`) simulating real API behaviour
3. Verifies apply succeeds

Without the fix this test fails with "Provider produced inconsistent result after apply". With the fix it passes.

Run with:
```
make unit TEST=./ec/ecresource/trafficfilterresource/...
```

Full suite (`make unit`) and linter (`make format lint`) also pass.

## Types of Changes

## Readiness Checklist

---

*This fix was developed with the assistance of [Claude](https://claude.ai) (Anthropic).*


dimuon (approved):
LGTM. Thanks for contributing!

PR #1008, opened 2026-06-19
Cross Project Search

tobio:
Changes:
- Adapt TrafficFilterRule.Source to *string
- Replace renamed status constants (ProjectStatusPhase*)
- Update OptionalMetadata patch tag handling
- Add linked project support with acceptance test
- Improve plan stability for computed metadata fields
- Rework the `linked` block to fix "Provider produced inconsistent result after apply" when linking a project, and to enable unlinking.

### Linked projects schema rework

The API-computed `status` is split out of the practitioner-controlled `linked.projects` map element into a top-level `statuses` Computed map, and `linked` is now Optional-only. This implements the Terraform Plugin Framework maintainers' "separate inputs from outputs" guidance: a `Computed` attribute nested inside a configured map element makes the element partially provider-owned, which previously caused a "Provider produced inconsistent result after apply" error when a new project was linked (status planned as `null` while the API returned `"enabled"`).

The schema is regenerated from `spec-mod.json` (via a new `modify_spec.sh` step) so the generated model types stay honest: `ProjectsValue` no longer carries a `Status` field, and `LinkedValue` only contains `Projects`.

### Unlinking

`expandLinkedProjectsForPatch` now emits an explicit `null` for keys that exist in prior state but are no longer in the plan. The serverless PATCH API treats an omitted `linked.projects` key as "no change", not "unlink", so the previous omission-based implementation silently left removed links intact and the target project could not be deleted.

### Cleanup

The now-unused `SetNullWhenEmptyString` plan modifier and the `patchLinkedStatusUseStateForUnknown` patch are removed; `statuses` is a plain Computed map and needs no plan modifier.

### Tests

Unit tests cover the unlink matrix (removed key, all keys removed, both empty, unknown state). `TestAcc_ElasticsearchProject_LinkedProjects` now exercises linking, a no-op plan, adding a second link, and unlinking.

Fixes #980


tobio:
Fixed lint: gofmt on 3 test files and removed unused strPtr. make lint and make unit now pass.

ec/internal/gen/serverless/resource_elasticsearch_project/elasticsearch_project_resource_gen.go
```diff
@@ -116,6 +116,53 @@ func ElasticsearchProjectResourceSchema(ctx context.Context) schema.Schema {
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"linked": schema.SingleNestedAttribute{
+				Attributes: map[string]schema.Attribute{
+					"projects": schema.MapNestedAttribute{
+						NestedObject: schema.NestedAttributeObject{
+							Attributes: map[string]schema.Attribute{
+								"status": schema.StringAttribute{
```
dimuon: Would it make sense to use `UseStateForUnknown`? Otherwise it produces this "known after apply" noise when other attribute gets changed. (`plan.Linked = useStateForUnknownOrNull(plan.Linked, state.Linked, resource_elasticsearch_project.NewLinkedValueNull())` doesn't help with this IIUC)

ec/ecresource/projectresource/elasticsearch.go
```diff
@@ -423,3 +448,85 @@ func (es elasticsearchApi) Delete(ctx context.Context, model resource_elasticsea
...
+		}
+	}
+
+	if len(projects) == 0 {
+		return nil
+	}
+	return &serverless.CreateLinkedRequest{Projects: projects}
+}
+
+func expandLinkedForPatchElasticsearch(plan resource_elasticsearch_project.ElasticsearchProjectModel) *serverless.OptionalLinkConfiguration {
```
dimuon: `expandLinkedForPatch*` only emits the projects present in the plan — it never emits a key that was in prior state but dropped from config. The API patches `linked.projects` with merge semantics (it's a `PATCH`, and `OptionalLinkedProject` / `OptionalLinkConfiguration` are `nullable: true`, i.e. you unlink by sending `{ "<id>": null }`), so omitting a removed key just leaves it linked. The result is that a user can't remove a single project from the map.

I confirmed this against the live API by extending `TestAcc_ElasticsearchProject_LinkedProjects` to link `{A, B}` and then re-apply with only `{A}`:

```
Error: Provider produced inconsistent result after apply
When applying changes to ec_elasticsearch_project.origin, provider produced an
unexpected new value: .linked.projects: new element "<B-id>" has appeared.
```

The PATCH dropped `B`, the API kept it, and the follow-up `Read` returned 2 links vs the planned 1 → hard apply failure.

There's a knock-on effect on teardown: the API refuses to delete a target while it's still linked (`projects.delete_project.project.in_use_by_linked_projects`), and since the provider can't emit the unlink, `terraform destroy` of a linked setup can fail and leave dangling projects.

This is exactly the case `optionalMetadataForTagPatch` already handles for tags — it diffs plan vs. state and emits `nil` for keys present in state but missing from the plan. I think `expandLinkedForPatch*` needs the same treatment: pass `state` in (the `Patch` method already has it) and set `projects[key] = nil` for each key in state but not in plan. Same applies to the observability and security expanders.

ec/ecresource/projectresource/metadata_tags.go
```diff
@@ -75,6 +76,43 @@ func patchMetadataTagsSchema(resp *resource.SchemaResponse) {
...
+			if stateField := stateCopy.FieldByName("SystemTags"); stateField.IsValid() {
+				field.Set(reflect.ValueOf(stateField.Interface().(basetypes.MapValue)))
+			}
+		}
+	}
+
+	return base.Interface().(T)
+}
+
+func metadataSystemTagsFromAPI(ctx context.Context, tags *serverless.ProjectSystemTags) (basetypes.MapValue, diag.Diagnostics) {
```
dimuon: Looks like the func ignores the args and always returns the same result.

dimuon:
**Test coverage for the new `linked` feature looks thin.**

Well covered: `optionalMetadataForTagPatch`, `product_tier` (unit + acc), and the `TrafficFilterRule.Source` → `*string` change.

Gaps worth filling:
- **No unit tests** for any of the new linked helpers — `flatten{Es,Obs,Sec}Linked`, `expandLinkedForCreate*`, `expandLinkedForPatch*`, `useStateForUnknownOrNull`. No unit test sets `Linked` on a model, so the new `linked` branches in `Create`/`Patch`/`Read`/`Modify` only ever exercise the `nil → null` path.
- `preserveMetadataForPlan`'s actual job — carrying `system_tags` from state into the plan — isn't asserted; the Modify tests just set `system_tags: MapNull` to keep compiling.
- Acceptance is a single happy path (`TestAcc_ElasticsearchProject_LinkedProjects`: ES → observability, create + empty-plan). No unlink/removal, multiple links, or observability/security as the origin.

Cheap, high-value additions: round-trip unit tests for `flatten*`/`expand*` (mirroring `expanders_test.go`, including the patch-removal case), a `preserveMetadataForPlan` test (state has `system_tags`, plan doesn't → preserved), and broadening the acc test to cover unlink + other origins.

dimuon (approved):
LGTM (one minor nit is that `metadataSystemTagsFromAPI` still returns null unconditionally)

PR #1012, opened 2026-06-24
Add serverless project sweepers for ACC leaks

dimuon:
## Summary
- Add an `ec_serverless_projects` acceptance test sweeper that deletes stale `terraform_acc_*` elasticsearch, observability, and security projects older than 3 hours.
- Run the serverless sweeper from the Buildkite `pre-exit` hook alongside the existing deployment sweeper (`-sweep-run=ec_deployments,ec_serverless_projects`).
- Increase the deployment sweeper stale window from 1h to 3 hours so both sweepers exceed the 120m ACC timeout and avoid cross-build pre-exit races.
- Fix inverted `CheckDestroy` fallback logic for all three serverless project resource types so manual cleanup works when Terraform destroy leaves resources behind.
- Harden serverless list pagination (`next_page` loop with duplicate-page / stale-token guards) and skip projects with a missing `created_at`.

## Test plan
- [x] `go test -c ./ec/acc/...` compiles
- [x] `Test_staleDeployment` passes with 3h window
- [x] Verify `make sweep SWEEPARGS='-sweep-run=ec_serverless_projects'` against a dev org
- [ ] Merge and confirm Buildkite acceptance `pre-exit` sweeps both deployments and serverless projects


dimuon (commented):
Overall this looks good — the CheckDestroy fix and CI sweeper wiring address the leaked-project problem well, and dropping the local cleanup script was the right call.

A few follow-ups below (pagination is the most important for the stated goal). Also, the PR title still mentions a cleanup script that was removed in the second commit — worth updating to something like **Add serverless project sweepers for ACC leaks**.

ec/acc/serverless_project_sweep_test.go
```diff
@@ -0,0 +1,224 @@
...
+			}(project.id)
+		}
+		wg.Wait()
+	}
+
+	return merr.ErrorOrNil()
+}
+
+func listElasticsearchProjects(ctx context.Context, client *serverless.ClientWithResponses) ([]serverlessProject, error) {
+	res, err := client.ListElasticsearchProjectsWithResponse(ctx, nil)
```
dimuon: `ElasticsearchProjectList` (and the observability/security equivalents) is paginated (`next_page` in the OpenAPI model), but we only process the first page here.

Given the motivating incident was hitting the **100-project org limit**, it's worth confirming the default page size. If the API returns a subset per page, this sweeper may miss leaked `terraform_acc_*` projects and not fully solve the problem.

Could we loop on `next_page` until exhausted (same for the other two list helpers)?

ec/acc/serverless_project_sweep_test.go
```diff
@@ -0,0 +1,224 @@
...
+			id:        string(item.Id),
+			name:      string(item.Name),
+			createdAt: item.Metadata.CreatedAt,
+		})
+	}
+	return projects, nil
+}
+
+func staleServerlessProject(createdAt time.Time) bool {
+	return createdAt.Before(time.Now().Add(-time.Hour))
```
dimuon: The deployment sweeper uses `lastModified` for staleness; here we use `createdAt`. That's probably fine for typical ACC runtimes, but a test running longer than 1h could theoretically have its project swept mid-run while still active.

`ProjectMetadata` doesn't expose `last_modified`, so this may be the best we can do — just flagging the asymmetry in case we want a longer safety margin or a different signal later.

ec/acc/serverless_project_sweep_test.go
```diff
@@ -0,0 +1,224 @@
...
+			if !staleServerlessProject(project.createdAt) {
+				continue
+			}
+
+			wg.Add(1)
+			go func(id string) {
+				defer wg.Done()
+				log.Printf("[DEBUG] Deleting serverless project %s", id)
+				if err := sweeper.delete(context.Background(), withResponses, id); err != nil {
+					merr = merr.Append(err)
```
dimuon, line 107: Minor: `merr.Append` is called from goroutines without synchronization. Same pattern as `deployment_sweep_test.go`, so consistent — but technically a data race under the race detector.

If we're touching this anyway, a `sync.Mutex` around append (or collecting errors on a channel) would be a small hardening.

ec/acc/serverless_project_sweep_test.go
dimuon: Good catch — updated in 7e53e68. `listServerlessProjects` now loops until `next_page` is empty, passing the token as a query param on subsequent requests (the generated client does not expose that parameter). For the ACC org today the API returns all projects in a single page, but the sweeper will handle larger orgs correctly.

ec/acc/serverless_project_sweep_test.go
dimuon: Agreed — kept the 1h `created_at` cutoff to match the deployment sweeper for now, and added a comment on `staleServerlessProject` documenting the asymmetry. Without `last_modified` in `ProjectMetadata` this is the best signal we have; we can revisit if ACC runs start exceeding the timeout more often.

ec/acc/serverless_project_sweep_test.go
dimuon: Done in 7e53e68 — `merr.Append` is now guarded by a `sync.Mutex` in the delete goroutines (and on list errors).

ec/acc/serverless_project_sweep_test.go
dimuon: Sanity-checked pagination when the ACC org had 98 projects: every type returned a single page with `next_page=""`, and `?page_size=1` / a synthetic `?next_page=` token did not change that. The leak was not caused by pagination — a single list call already returned everything. `next_page` is only in the response schema (deprecated) in our OpenAPI and on the published API docs, not as a query parameter.

Pushed `427f949` with guards so a broken pagination implementation fails loudly: duplicate project IDs across pages, a stale `next_page` token, or >100 pages all return an error instead of silently re-processing page 1.

.buildkite/hooks/pre-exit
```diff
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 if  [[ "$BUILDKITE_STEP_KEY" == "acceptance-tests" ]]; then
-  echo "--- Sweeps any deployments older than 1h."
-  EC_API_KEY=$TERRAFORM_PROVIDER_API_KEY_SECRET SWEEPARGS=-sweep-run=ec_deployments make sweep
+  echo "--- Sweeps any deployments older than 1h and serverless projects older than 3h."
+  EC_API_KEY=$TERRAFORM_PROVIDER_API_KEY_SECRET SWEEPARGS='-sweep-run=ec_deployments -sweep-run=ec_serverless_projects' make sweep
```
gigerdo: **[AI comment]** Passing `-sweep-run` twice does not combine the filters. It is a single `flag.String` in terraform-plugin-testing, so Go's flag parser keeps only the **last** value (`ec_serverless_projects`), and the existing `ec_deployments` sweep silently stops running. Use one comma-separated value instead:

```sh
SWEEPARGS='-sweep-run=ec_deployments,ec_serverless_projects'
```

ec/acc/serverless_project_sweep_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	baseURL := strings.TrimRight(cfg.Host, "/") + "/api/v1/serverless/projects/" + projectType
+	projects := make([]serverlessProject, 0)
+	seenIDs := make(map[string]struct{})
+	nextPage := ""
+
+	for pageNum := 0; pageNum < maxServerlessProjectListPages; pageNum++ {
+		reqURL := baseURL
+		// next_page is returned in list responses but is not declared as a query
+		// parameter in the checked-in OpenAPI spec (or the published API docs).
```
gigerdo, line 136: **[AI comment]** This rationale does not hold for the checked-in client. `next_page`/`page_size` are declared as `in: query` parameters in `serverless-project-api-dereferenced.yml`, and `client.gen.go` exposes them via `ListElasticsearchProjectsParams{PageSize, NextPage}` plus `ElasticsearchProjectList{Items, NextPage}`. The entire raw-HTTP `listServerlessProjects` (manual URL building, auth, JSON) can be replaced with the typed `ListXxxProjectsWithResponse` client you already type-assert, removing the second client construction and duplicate auth path. (`next_page` is only absent from `spec.json`/`spec-mod.json`, not from the actual generation source.)

ec/acc/serverless_project_sweep_test.go
```diff
@@ -0,0 +1,267 @@
...
+		nextPage = *page.NextPage
+	}
+
+	return nil, fmt.Errorf("listing %s projects: exceeded %d pages", projectType, maxServerlessProjectListPages)
+}
+
+// staleServerlessProject uses created_at because ProjectMetadata does not
+// expose last_modified (unlike deployment sweepers).
+func staleServerlessProject(createdAt time.Time) bool {
+	return createdAt.Before(time.Now().Add(-serverlessProjectStaleAfter))
```
gigerdo, line 209: **[AI comment]** Deletion risk: if the API omits or zeroes `created_at`, `createdAt` is the zero `time.Time`, which is always `Before(now-3h)`, so the project is treated as stale and deleted. That defeats the 3h safety window for a concurrently-running build's fresh projects. Guard the zero value:

```go
if createdAt.IsZero() {
    return false
}
```

gigerdo (commented):
[AI comment] Automated review notes from an AI assistant, limited to the critical items. Please verify before acting.

ec/acc/serverless_project_sweep_test.go
dimuon: Thanks for flagging this — I re-checked the generation source on `aee7967`.

The HTTP client is generated from `serverless-project-api-dereferenced.yml` via `oapi-codegen` (`serverless.go`), not from `spec.json` / `spec-mod.json` (those are only used for Terraform resource codegen).

On that YAML, `GET /api/v1/serverless/projects/{type}` only declares `traffic_filter`, `linked`, and `tags` as query parameters. `next_page` appears only on the response schemas (`ElasticsearchProjectList`, etc.), marked deprecated — there is no `name: next_page` / `name: page_size` with `in: query` on the list operations.

That matches the generated client: `ListElasticsearchProjectsParams` only has `TrafficFilter`, `Linked`, and `Tags`. `NextPage` exists on the response type (`ElasticsearchProjectList`), and `NewListElasticsearchProjectsRequest` only serializes those three query params — so the typed client cannot pass a pagination token today without extending the upstream OpenAPI and regenerating.

`spec-mod.json` also has no `next_page` / `page_size` references, but that is expected since it is not the client input anyway.

Given that, the raw-HTTP `listServerlessProjects` helper is intentional: it follows `next_page` from the response and passes `?next_page=` on subsequent requests. Happy to switch to the typed client once the upstream spec exposes pagination as query params and we regenerate — until then this is the only way to page from ACC.

.buildkite/hooks/pre-exit
dimuon: Good catch — fixed in `aee7967`. `SWEEPARGS` now uses a single comma-separated value:

```sh
SWEEPARGS='-sweep-run=ec_deployments,ec_serverless_projects'
```

Confirmed against `terraform-plugin-testing@v1.16.0`: `-sweep-run` is one `flag.String` parsed as comma-separated; passing the flag twice only kept the last value.

ec/acc/serverless_project_sweep_test.go
dimuon: Agreed — fixed in `aee7967`. `staleServerlessProject` now returns `false` when `createdAt.IsZero()` so projects with a missing `created_at` are not treated as stale.

ec/acc/serverless_project_sweep_test.go
gigerdo: **[AI comment]** Correction / retraction — you're right, and my earlier comment was wrong. I had reviewed against a locally-divergent checkout (279 commits ahead of `origin/master`) whose regenerated serverless client happened to include `page_size`/`next_page` query params. On this PR branch and on current upstream `master`, `ListElasticsearchProjectsParams` only exposes `TrafficFilter`/`Linked`, the dereferenced YAML doesn't declare pagination as `in: query`, and `next_page` exists only on the (deprecated) `ElasticsearchProjectList` response type. So the typed client genuinely can't page today, and the raw-HTTP `listServerlessProjects` is the right call until the upstream spec exposes pagination and the client is regenerated. Thanks for verifying so thoroughly.

dimuon:
@gigerdo , Could you take another look? I've addressed the concerns you raised.

gigerdo (approved):
LGTM :+1:

gigerdo (approved):

PR #1017, opened 2026-07-03
feat(ec_deployment): add rolling_zone strategy for AZ-aware rolling restarts

scathatheworm:
## Summary

Adds a new `rolling_zone` strategy value to the `elasticsearch.strategy` attribute in `ec_deployment`. When set, the Elastic Cloud API performs a rolling restart grouped by availability zone (`group_by: logical_zone_name`), ensuring at most one zone is restarted at a time.

### Changes

- New `rolling_zone` strategy option maps to `group_by: logical_zone_name` in the API's `RollingStrategyConfig`
- `ModifyPlan` now validates that `rolling_zone` is not used during a major version upgrade (the API requires `group_by: __all__` for major upgrades); an explicit error is returned instead of silently rewriting the strategy
- Schema description updated with a note on the major-version constraint
- Docs regenerated via `make docs-generate`

### Existing strategies

| Terraform value | API `group_by` | Behaviour |
|---|---|---|
| `rolling_all` | `__all__` | All nodes restart together |
| `rolling_zone` | `logical_zone_name` | One AZ at a time (this PR) |


scathatheworm:
I have signed the Contributor Agreement

dimuon:
Hi @scathatheworm — thanks for signing the CLA.

The check is still failing because it matches the **email in your commits**, not just your GitHub username. Both commits use `david.ponessa@booking.com`, and GitHub doesn't attribute them to your account (`author` is null), which usually means that email isn't a verified email on your GitHub profile and/or isn't the one on your CLA record.

Please use the **same email** for all three: git `user.email`, a verified GitHub email, and the CLA signature. Then rebase your commits and force-push. If you're unsure which email is on your CLA, check your signing confirmation or email [cla-questions@elastic.co](mailto:cla-questions@elastic.co) (listed on the [contributor agreement page](https://www.elastic.co/contributor-agreement)).

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
```diff
@@ -194,6 +196,7 @@ func (plan DeploymentTF) UpdateRequest(ctx context.Context, client *api.API, sta
 		// can't be full once the cluster has been created, so the Strategy must be set
 		// to "partial".
 		ensurePartialSnapshotStrategy(elasticsearchPayload)
+		ensureMajorVersionStrategy(elasticsearchPayload, state.Version.ValueString(), plan.Version.ValueString())
```
dimuon: `ensureMajorVersionStrategy` silently rewrites `group_by` from `logical_zone_name` to `__all__` in the update payload, while Terraform config/state still has `strategy = "rolling_zone"`.

**Why this is problematic**

1. **Plan/state vs API mismatch** — `strategy` is not read from the API's `transient` block; `read.go` re-injects it from plan/state. After apply, Terraform still reflects `rolling_zone`, but the API ran with `__all__` (`rolling_all` semantics).

2. **Can't fix this in `ModifyPlan` by rewriting** — For a user-set `Optional` attribute, the framework requires preserving the exact config value (or prior state). Silently changing `rolling_zone` → `rolling_all` there would violate that contract unless `strategy` were `Computed`, which doesn't fit this use case.

3. **`ValidateConfig` isn't an option** — It only has config, not state, so it can't detect a major version upgrade (`state.version` vs `plan.version`).

**Suggested fix**

- Remove this call (and `ensureMajorVersionStrategy`).
- Validate in `ModifyPlan` instead (where we already have plan + state, e.g. alongside `UpdateDedicatedMasterTier`): fail when `strategy == "rolling_zone"` and version crosses a major boundary.
- Use `semver.Parse` and compare `.Major` (as in `node_roles.go`), not `strings.SplitN`.
- Example error:
  > `rolling_zone` cannot be used for major version upgrades. Set `strategy = "rolling_all"` when upgrading across a major version boundary (API requires `group_by: __all__`).

ec/ecresource/deploymentresource/elasticsearch/v2/schema.go
```diff
@@ -45,11 +45,13 @@ const (
 	strategyGrowAndShrink        = "grow_and_shrink"
 	strategyRollingGrowAndShrink = "rolling_grow_and_shrink"
 	strategyRollingAll           = "rolling_all"
+	strategyRollingZone          = "rolling_zone"
 )
 
 // List of update strategies availables.
 var strategiesList = []string{
 	strategyAutodetect, strategyGrowAndShrink, strategyRollingGrowAndShrink, strategyRollingAll,
+	strategyRollingZone,
```
dimuon, line 54: The schema section in `docs/resources/deployment.md` is generated by `make docs-generate` from the provider schema — editing the markdown directly won't stick.

Please extend the `strategy` attribute `Description` (around line 152) to document that `rolling_zone` cannot be used for major version upgrades and that `rolling_all` is required instead (API requires `group_by: __all__` per `RollingStrategyConfig` in the Elastic Cloud API).

You can follow the same pattern as `snapshot_source`, which includes a `~> **Note on behavior**` block in its `Description` in this file. After updating the description, run `make docs-generate` so the constraint appears in the published docs.

dimuon:
Thanks for contributing this — `rolling_zone` is a useful addition and the implementation is clean.

A small consistency issue between the PR title/description and the actual diff:

- The **title** says *"add rolling and rolling_zone"*, but the diff only adds **`rolling_zone`**.
- The **description** documents a `rolling` value (`group_by: __name__`) and says it has *"same behaviour as the existing `rolling_all`*, but that's not accurate — **`rolling_all` maps to `__all__`** (all nodes restart together), not node-by-node rolling. Node-by-node (`__name__`) is the API default and isn't exposed as a named strategy today.

Could you align the title and description with what's actually in the PR? Either drop `rolling` from the title/description, or add it in a follow-up if that's still intended.

dimuon:
Thanks again for the PR.

One housekeeping item: there's no `.changelog/1017.txt` in the PR. [CONTRIBUTING.md](https://github.com/elastic/terraform-provider-ec/blob/master/CONTRIBUTING.md) requires a changelog entry for user-facing changes — please add one before merge (see [example entries](https://github.com/elastic/terraform-provider-ec/tree/master/.changelog)).

scathatheworm:
Thanks for the review @dimuon ! I believe I have a addressed all your review items in the latest commit.

ec/ecresource/deploymentresource/deployment/v2/deployment_update_payload.go
scathatheworm: I've addressed these in https://github.com/elastic/terraform-provider-ec/pull/1017/changes/8804d87bc0ee7bfd3cbc28db0266ab68d13105aa can you check?

dimuon (approved):
Thanks for contributing!

scathatheworm:
Hi @dimuon , just to know, when is a new version that includes this going to be released?

dimuon:
Hi @scathatheworm , Unfortunately, I can’t provide an exact date right now, but we’re working on improving our release cadence so we can deliver updates more frequently.

Issue #906, opened 2025-05-09
ec_stack data provider is not able to retrieve stack 7.17 version


kunisen:
## Description

- Using terraform ec provider and specifying 7.17 version in ec_stack cannot deploy 7.17 version. 
- Specifying version directly in `resource "ec_deployment"` block can work

## Details

error detail

```bash
$ terraform apply
data.ec_stack.latest: Reading...
╷
│ Error: failed to obtain a stack version matching "7.17.28": please specify a valid version_regex
│ 
│   with data.ec_stack.latest,
│   on deployment.tf line 18, in data "ec_stack" "latest":
│   18: data "ec_stack" "latest" {
│ 
│ failed to obtain a stack version matching "7.17.28": please specify a valid version_regex
```

terraform & ec version

```
terraform -v 
Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/elastic/ec v0.12.2
```

deployment.tf:

```tf
terraform {
  required_version = ">= 0.12.29"

  required_providers {
    ec = {
      source  = "elastic/ec"
      version = "0.12.2"
    }
  }
}

provider "ec" {
  # my test trial user (not elastic account)
  apikey = "foobar"
}


# Retrieve the latest stack pack version
data "ec_stack" "latest" {
  version_regex = "7.17.28"
  region        = "us-west-1"
}

# Create an Elastic Cloud deployment
resource "ec_deployment" "example_minimal" {
  # Optional name.
  name = "kuni-test-7.17.28"

  region                 = "us-west-1"
  version                = data.ec_stack.latest.version
  deployment_template_id = "aws-storage-optimized"

  elasticsearch = {
      hot = {
        autoscaling = {}
    }
  }

  kibana = {}
}
```

## Workaround

If we instead specify version in `resource "ec_deployment"` block directly, 

```tf
  region                 = "us-east-1"
  version                = "7.17.28"
  deployment_template_id = "aws-storage-optimized"
```

it can deploy

```
terraform apply

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # ec_deployment.example_minimal will be created
  + resource "ec_deployment" "example_minimal" {
      + alias                  = (known after apply)
      + deployment_template_id = "aws-storage-optimized"
      + elasticsearch          = {
          + autoscale      = (known after apply)
          + cloud_id       = (known after apply)
          + hot            = {
              + autoscaling                           = {
[...]

PR #1043, opened 2026-09-03
allow matching deleted stack versions if explicit version is passed

myrcutio:
## Description
If a stack version is deleted, existing stacks on that version are still valid.  As described in #906 the current behavior raises an error if a configuration is pinned to a deleted version, even for operations such as plan.  Related to this, explicitly pinned versions (such as 8.19.2 for example) which are substrings of newer versions (e.x. 8.19.21) should not greedy match.

## Related Issues
#906 

## Motivation and Context
This allows plan and apply operations to be conducted on stacks without modifying the cluster version.

## How Has This Been Tested?
Tested locally on cluster with version 9.5.2, and locally on a cluster with version 8.19.2.

## Types of Changes
- [x] Testing

## Readiness Checklist


gigerdo (commented):
Superseded by the review with inline comments below.

ec/ecdatasource/stackdatasource/datasource.go
```diff
@@ -68,14 +68,16 @@ func (d DataSource) Read(ctx context.Context, request datasource.ReadRequest, re
 		return
 	}
 
+	re, _ := regexp.Compile("\\^?[0-9]\\.[0-9]*\\.[0-9]*\\$?")
 	res, err := stackapi.List(stackapi.ListParams{
-		API:    d.client,
-		Region: newState.Region.ValueString(),
+		API:     d.client,
+		Region:  newState.Region.ValueString(),
+		Deleted: re.MatchString(newState.VersionRegex.ValueString()),
```
gigerdo: Requesting `show_deleted` only for an exact pin is the right idea: 9.5.2 is soft-deleted, so the default list will never return it, but keep-on-current is still valid.

This detector is too loose (`8..2` matches) and too tight (major is a single `[0-9]`, so `10.0.0` will not). Ignore-error compile is also easy to miss.

Parse a concrete semver, optionally wrapped in `^…$` (`9.5.2` / `^9.5.2$`). Leave `latest`, `9.5.?`, and `9\.5\..+` on `show_deleted=false`.

ec/ecdatasource/stackdatasource/datasource.go
```diff
@@ -139,7 +141,7 @@ func stackFromFilters(expr, version string, locked bool, stacks []*models.StackV
 		return stacks[0], nil
 	}
 
-	re, err := regexp.Compile(expr)
+	re, err := regexp.Compile(`(?:` + expr + `)\b`)
```
gigerdo, line 185: This rewrites every user regex, not just exact pins. `7.8.?` becomes `(?:7.8.?)\b`, and `^8.19.2$` becomes `(?:^8.19.2$)\b`.

The 8.19.2 vs 8.19.21 case is real, but it should be an exact match (quote/anchor when the expression is a concrete version), not a blanket `\b` suffix.

ec/ecdatasource/stackdatasource/datasource_test.go
```diff
@@ -255,6 +257,11 @@ func Test_stackFromFilters(t *testing.T) {
 			args: args{expr: "7.9.0", stacks: stackPacks},
 			want: &models.StackVersionConfig{Version: "7.9.0"},
 		},
+		{
+			name: "returns the stack pack without greedy regex",
+			args: args{expr: "8.19.2", stacks: stackPacks},
+			want: &models.StackVersionConfig{Version: "8.19.2"},
+		},
```
gigerdo, line 264: Good coverage for the 8.19.2 / 8.19.21 case. Keep this test; I would make it pass by treating `8.19.2` as an exact version (quoted/anchored) rather than wrapping all expressions with `\b`.

gigerdo (commented):
Thanks for this — it is the right shape for #906. Requesting `show_deleted` only when the caller already named an exact version is a resolve, not a catalog of withdrawn packs. That is better than synthesizing an empty stack object (we closed #1045 for that reason).

Code notes are on the snippets below.

Also before merge:
- Add `.changelog/1043.txt` and a short `ec_stack` note: an exact pin can resolve a withdrawn version so no-op plans keep working; `latest` still follows the advertised newest version.
- `latest` cannot keep 9.5.2 across a yank-without-replacement. Plugin-framework data sources do not receive prior state, so `lock` cannot remember the last apply. Pin `9.5.2` until 9.5.3 is advertised.
- CLA is failing on `5c7f06a` — needs a signed Elastic CLA.

ec/ecdatasource/stackdatasource/datasource_test.go
dimuon, line 264: Thanks for adding the 8.19.2 / 8.19.21 case — that is a real matching bug.

These cover matching once the list is already in hand. Could we also cover **when** we set `show_deleted`?

A small table for “is this an exact pin?” would lock that down:
- `9.5.2` / `^9.5.2$` / `10.0.0` → true
- `latest`, `9.5.?`, `9\.5\..+` → false

myrcutio: Had to switch the approach a bit, it seems that `show_deleted` behavior changed since last week and no longer shows deleted versions when the API is authenticated.  I did move the "is this an exact version?" logic into a function with coverage, so it's in keeping with the spirit there.

ec/ecdatasource/stackdatasource/datasource.go
myrcutio: I went back and forth on this one, but i think wrapping with boundary is less intrusive than trimming and adding anchors back.  I don't think there's a valid semver that includes a word boundary, and this still allows the user to specify anchors to their hearts content while still fixing the `.2 ~ .21` case

ec/ecdatasource/stackdatasource/datasource.go
myrcutio: good catch on the 10 and .. cases!  I swapped it out for `[0-9]+` on each and added a panic on the compile error.  it's a static regex so it shouldn't ever trigger anyway.

ec/ecdatasource/stackdatasource/datasource.go
gigerdo: Now with the recent changes is this still necessary? Since now a 8.19.2 would go into `isExactVersion`

ec/ecdatasource/stackdatasource/datasource.go
```diff
@@ -68,25 +71,56 @@ func (d DataSource) Read(ctx context.Context, request datasource.ReadRequest, re
 		return
 	}
 
-	res, err := stackapi.List(stackapi.ListParams{
-		API:    d.client,
-		Region: newState.Region.ValueString(),
-	})
+	var stackVersion *models.StackVersionConfig
+	exact, err := isExactVersion(newState.VersionRegex.ValueString())
```
gigerdo, line 75: One thing I wonder here: This fixes any user set on a specific version. But when using a specific version, why even use `ec_stack`? The main use-case for `ec_stack` is to dynamically choose a version.

One further caveat is: If a version is pulled/soft-deleted it can still be used in already existing deployments. However no new deployment can use that version. So returning a specific version from ec_stack would work, but it would then fail in any new ec_deployment resource.

ec/ecdatasource/stackdatasource/datasource.go
myrcutio: nope, no longer needed.  removed.

ec/ecdatasource/stackdatasource/datasource.go
myrcutio: It's an ambiguity, but i think probably the best we can do with the current state of the API.  If a user tries to specify a deleted version and it fails on apply i think that's a reasonable behavior.

gigerdo: To rephrase the question: Is there any reason for a user to use `ec_stack` when specifying an explicit stack version?

PR #965, opened 2026-01-12
Add support for remote_cluster traffic filter type

ryunosuke121:
## Description
   This PR adds support for the  traffic filter type to the traffic filter resource and data source. This allows users to configure traffic filters for remote cluster connections.

   ## Related Issues
   https://github.com/elastic/terraform-provider-ec/issues/966

   ## Motivation and Context
   Elastic Cloud supports remote cluster traffic filters to control access between clusters. This enhancement allows Terraform users to manage these remote cluster traffic filter rules through the provider.

   ## How Has This Been Tested?
   - Added unit tests for the new  and  fields
   - Updated expanders and flatteners tests to cover the new fields
   - All existing tests continue to pass

   ## Types of Changes
   - [x] New feature (non-breaking change which adds functionality)

   ## Readiness Checklist
   - [x] My code follows the code style of this project
   - [ ] My change requires a change to the documentation
   - [ ] I have updated the documentation accordingly
   - [x] I have added tests to cover my changes
   - [x] All new and existing tests passed

Issue #947, opened 2025-11-28
Expose ingest endpoint as part of ec_deployment_resource/integrations_server


simitt:
## Overview

As part of adding the `ingest` endpoint to the Integrations Server in ECH, the terraform provider should also return the `ingest` endpoint similar to how `apm` endpoint is returned. 

## Possible Implementation

Expose the `integration` endpoint similar to https://github.com/elastic/terraform-provider-ec/blob/master/ec/ecresource/deploymentresource/integrationsserver/v2/schema.go#L82

## Testing

At least unit tests

## Context

We are adding the `ingest` endpoint as an alternative to `apm` in Elastic Cloud Hosted. For full support, the endpoint should be also supported with the terraform provider. 

## Blockers
This is currently blocked until the `ingest` endpoint is actually available via the Cloud API.

PR #1049, opened 2026-09-08
[1.3] OpenSpec authoring conventions + exemplar specs

dimuon:
## Summary
- Add `dev-docs/high-level/openspec-requirements.md`: Purpose / SHALL-MUST requirements / Given-When-Then scenarios, `openspec/specs/` vs `openspec/changes/`, and the cloud-provider rules (no Docker stack; agents never run `TF_ACC`; changelog is `.changelog/{PR}.txt`; going forward only new/changed behavior needs a spec).
- Seed two **current-behavior** canonical specs as copy examples: `openspec/specs/deployment-traffic-filter/` (`ec_deployment_traffic_filter` resource) and `openspec/specs/stack/` (`ec_stack` data source).
- Point `openspec/config.yaml`, `AGENTS.md`, and the high-level dev docs at the conventions.

Relates to elastic/cp-hosted-team#4365 (epic elastic/cp-hosted-team#4362).

## Divergences from the ticket
- **No spec for issue terraform-provider-ec#966.** `remote_cluster` already shipped in #965; the traffic-filter spec documents that as current behavior, not as work to implement.
- **No integrations-server / terraform-provider-ec#947 exemplar.** `#947` is unimplemented (`ingest`); that contract belongs in `openspec/changes/` when 1.4/1.5 implement it, not in `openspec/specs/`.
- **Direct to `openspec/specs/`.** These exemplars capture existing code. New/changed behavior still goes through `openspec/changes/` then `openspec archive`.

## Test plan
- [x] `make check-openspec` locally (2 specs passed)
- [ ] Confirm `OpenSpec CI / Validate OpenSpecs` is green on this PR
- [ ] Human review of the conventions doc (Purpose / SHALL-MUST / Scenarios)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,6 +16,7 @@ and related resources through the Elastic Cloud API.
 - Testing (unit + acceptance) and required env: [`testing.md`](./dev-docs/high-level/testing.md)
 - Generated clients (serverless OpenAPI) and regeneration: [`generated-clients.md`](./dev-docs/high-level/generated-clients.md)
 - Documentation generation (`tfplugindocs`): [`documentation.md`](./dev-docs/high-level/documentation.md)
+- OpenSpec authoring (Purpose / SHALL-MUST / Scenarios): [`openspec-requirements.md`](./dev-docs/high-level/openspec-requirements.md)
 
 ## Testing note — acceptance tests hit the real, paid Elastic Cloud API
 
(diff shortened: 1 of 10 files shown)


gigerdo (commented):
Keeping the `SHALL`s as current behavior is right — a later lock fix would just be a change that updates the spec. Encoding the bug as a copy-me requirement is the unfortunate part.

Please label it as a **Known gap** so agents do not treat “lock does nothing” as the desired contract.

**`openspec/specs/stack/spec.md`** — after Requirements:

```markdown
## Known gaps (non-normative)

These record limitations of the current implementation. A later fix is a change that
updates the `SHALL`s above; do not treat this section as the desired contract.

### `lock` does not pin

Schema and registry docs describe `lock` as pinning the `latest` result so a new
stack release does not cascade. `stackFromFilters` still implements that pin when
`version` is non-empty, and the unit test `"returns the latest stackpack with a
locked version"` expects it. Read loads configuration only, so `version` is empty
and the pin is not Terraform-observable. A change that makes `lock` work should
replace the Read requirement above.

### `latest` with an empty list

The `latest` path indexes the first list element with no length check. An empty
list panics; it does not use the regex no-match diagnostic. A change should add
an explicit error (or documented empty handling) and a scenario with a positive
THEN.
```

Then drop the “Schema copy describes…” paragraph from the lock requirement (that belongs in Known gaps). Same for “An empty list is unhandled…” in the `latest` requirement.

**`openspec-requirements.md`** — one sentence under Authoring so the next exemplar copies the pattern:

> When a current-behavior spec captures a known limitation, keep the `SHALL` as the observable rule and add a non-normative **Known gaps** note. A later bug fix is a change that updates the `SHALL`.

<small>*(AI-authored via Cursor 🤖)*</small>

gigerdo (approved):
Looks good. Conventions and the two exemplars match current behavior; the Known gaps note above is a non-blocking follow-up.

<small>*(AI-authored via Cursor 🤖)*</small>

PR #1055, opened 2026-09-14
test(acc): pin Kibana to 2g in migrate-to-latest-hw test to avoid OOMs

dimuon:
## Description

`TestAccDeployment_migrate_to_latest_hw` used `kibana = {}`, so Kibana fell through to the deployment-template default of **1 GB**. That default is insufficient: Kibana instances fail to start, which fails the test during create.

This sets an explicit `size = "2g"` on both fixtures (create + migrate) and updates the size assertions. Same approach as #1024 and #1025.

## Motivation and Context

[Acceptance build 2991](https://buildkite.com/elastic/terraform-provider-ec-acceptance/builds/2991) failed on `ec_deployment.cpu_optimized` with:

```
found deployment plan errors: 1 error occurred:
    * deployment [91519ef6d7208c7c1571e871e4d30808] -
    [kibana][3bbe1a7eefbd40c7a15ab2dee9cb0b4e]: caught error: "Plan change
    failed: Some instances were not running"
```

#1024 called this test out of scope; this is that follow-up.

## How Has This Been Tested?

- Test-only fixture + assertion change; no provider behavior change.
- Did not run `TF_ACC=1` locally (agents never run acceptance tests). Confirmation is the Buildkite acceptance pipeline on this PR.

## Types of Changes
- [ ] Documentation

## Readiness Checklist

Made with [Cursor](https://cursor.com)


beiske (approved):
Didn't we change the template default to 2GB a while back?

dimuon:
Some older tests can still explicitly set Kibana's size to 1 GB, like in this case.

dimuon:
Actually, @beiske, you're right - the test doesn't set the size explicitly. Instead, the provider has some really old logic that bumps the size to 1 GB for Kibana if the config doesn't set it explicitly. There's similar logic for the zone count too.

dimuon:
I'll address this.

dimuon:
https://github.com/elastic/terraform-provider-ec/pull/1057

